PR #11, opened 2022-05-05
disabled eslint is no longer prone to not being active when file gets formatted

meszaros-lajos-gyorgy:


frostburn:
I'd prefer
```
  rules: {
    "no-constant-condition": 0,
     ....
```
In `.eslintrc.cjs`

frostburn:
I'll fix this in my upcoming `mos` PR.

Issue #2, opened 2022-05-01
Scale Generation


frostburn:
Implement scale generation for producing plain text scales and initial states for non-destructive scale editing.
### Old
- [x] Equal Temperament: Number of divisions, Interval to divide
- [x] Rank 2: Same as Scale Workshop 1
- [x] Harmonic series segment: Lowest harmonic, Highest Harmonic, Step
- [x] Subharmonic segment: Similar to harmonic segment
- [x] Enumerate chord: Same as SW1
- [x] Combination product set: Same as SW1 but with configurable equave
### New
- [x] MOS for ET: Number of large steps, number of small steps, size of large step, size of small step, mode, equave
- [ ] Maximum variety for ET: Example of MV3: 3L 4s x 5L 2s = 4 3 2 4 3 4 2 (Nicetone)
- [x] Color temperament: Similar to rank 2, but the generator and period is inferred from a comma names such as _Sagugu & Ru_. TODO: Figure out how to deal with other ranks
- [x] Named temperament: Same as above, but there's an internal dictionary that converts names like _Ripple_ to color commas (_Quingu & Ruru_ in this case). TODO: Figure out how to deal with limit. _Ripple_ in the 5-limit is simply _Quingu_.
- [x] More Just Intonation shapes like cubes and octahedra. Option to stack them in monzo space if stackable.
- [ ] Generator offset scales (Covers stuff like stacked alternating 1\11 and 1\13)
- [ ] Fokker blocks
- [ ] Ringer scales

frostburn:
Moved the complicated stuff to separate issues.

PR #14, opened 2022-05-07
Semi-convergents, MOS and temperaments

frostburn:
Refs #2


frostburn:
It is very likely that I will drop the MOS and temperament code from this repository and implement them as separate packages. If whoever chooses to review this is short on time, let me know and I'll remove the code that will not live in SW2 long term.

PR #21, opened 2022-05-08
Implement exporters for Scala and AnaMark

frostburn:


frostburn:
There's some inconsistencies between the APIs of different exporters which needs to be unified.

frostburn:
Approved via Discord.

PR #46, opened 2022-06-04
Implement tempering mappings

frostburn:
Add temperaments as a dependency


frostburn:
Now that we have modals in `main`, I guess I could implement the related scale creation/modification modals in this branch as well.

frostburn:
Tempering overlaps with the MOS modal I'm currently working on. Let's put this PR on hold until I've completed that and we have it merged. There's stuff about MOS patterns and modes that I would like to include in these modals as well where applicable. Most notably the rank-2 generator method should be expanded to include the number of periods and the vals/commas methods should simply fill that in.

frostburn:
The tempering modifier should have a direct method that presents the mapping to the user before applying.

frostburn:
This will go through major changes. Closing PR for now.

PR #79, opened 2022-06-20
Implement importers for Scala .scl and Anamark .tun

frostburn:


frostburn:
The Korg importer is broken. I will separate it to another branch.

frostburn:
I will figure out what's going on with Korg later. This PR with only Scala and Anamark is good for review again.

SeanArchibald (approved):

PR #85, opened 2022-06-21
Add a button for calculating the base frequency from MIDI note

frostburn:
refs #52

### DIFF
--- a/src/components/ScaleBuilder.vue
+++ b/src/components/ScaleBuilder.vue
@@ -1,8 +1,8 @@
 <script setup lang="ts">
-import { computed } from "vue";
+import { computed, ref } from "vue";
 import TuningTable from "./TuningTable.vue";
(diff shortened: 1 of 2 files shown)


src/components/ScaleBuilder.vue
```diff
@@ -154,9 +165,17 @@ const updatebaseMidiNote = debounce((event: Event) => {
         </div>
 
         <div class="control">
+          <button
+            @click="autoFrequency"
+            style="position: absolute; right: 1.4rem; z-index: 9001"
```
frostburn, line 170: I know this isn't good CSS. @SeanArchibald you will need to help me here.

SeanArchibald: This is a good solution for now and I might end up changing the position of the button in future anyway which will make things simpler

SeanArchibald (approved):

Issue #89, opened 2022-06-22
Composite Linetype


frostburn:
`ExtendedMonzo.toScaleLine` can produce composite intervals. These need to be parseable.

Issue #88, opened 2022-06-22
Monzo Linetype


frostburn:
`ExtendedMonzo.toScaleLine` can produce monzos in extreme scenarios. These need to be parseable.

Issue #26, opened 2022-05-09
Generalized N-of-EDO Linetype


frostburn:
Implement line type for general equal temperament `n\m<p/q>`: `n` steps of `p/q` divided into `m` equal steps.

Also consider adding `edo`, `edf` and `edt` postfixes e.g. `7\13edt`: seven steps of the tritave divided into 13 equal steps.

PR #90, opened 2022-06-22
Extend parser

frostburn:
Improve representation of negative composite offsets in monzo revers parsing
Implement generalized N-of-EDO line type
Implement monzo line type
Implement composite line type

refs #26 #88 #89


SeanArchibald (approved):
Very useful

PR #91, opened 2022-06-22
Implement automatic key colors based on scale size

frostburn:
refs #53

### DIFF
--- a/src/__tests__/util.spec.ts
+++ b/src/__tests__/util.spec.ts
@@ -1,6 +1,6 @@
 import { describe, it, expect } from "vitest";
 
-import { formatExponential, formatHertz } from "../utils";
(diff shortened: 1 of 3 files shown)


SeanArchibald (commented):
I approve in principal about the changes here, though one example I tested was 17ed2 and the result was:

white white black white white white white white white white black white white white white white white

It seems a really weird key layout for me with 15 white notes and 2 black notes so I'm not sure if this was intended.

frostburn:
Such extreme scales are not intended. I'll check if using fifth-based scales helps.

frostburn:
Changed to fifth/fourth based construction. Mavila edos still get only two black keys (as they should IMHO), but 17 gets more now.

SeanArchibald:
Works well, looks good to merge

PR #92, opened 2022-06-22
Apply key colors to tuning table rows

frostburn:


SeanArchibald (approved):

SeanArchibald:
Looks like the colours start on note 0 instead of the reference MIDI note number defined by the user

frostburn:
Fixed the starting offset for key colors.

Issue #73, opened 2022-06-20
Polyphonic and configurable MIDI


frostburn:
- [x] Channel set selection
- [x] Velocity sensitivity on/off
- [x] White keys mode

PR #103, opened 2022-06-24
Make MIDI output polyphonic

frostburn:
Make it possible to select MIDI input channels

refs #73

### DIFF
--- a/src/views/MidiView.vue
+++ b/src/views/MidiView.vue
@@ -5,9 +5,16 @@ import { Input, Output, WebMidi } from "webmidi";
 const props = defineProps<{
   midiInput: Input | null;
   midiOutput: Output | null;
+  midiInputChannels: Set<number>;
+  midiOutputChannels: Set<number>;
 }>();
 
-const emit = defineEmits(["update:midiInput", "update:midiOutput"]);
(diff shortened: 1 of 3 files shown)


src/views/MidiView.vue
```diff
@@ -75,6 +104,18 @@ onMounted(async () => {
               {{ (input.manufacturer || "(Generic)") + ": " + input.name }}
             </option>
           </select>
+          <div class="control">
```
frostburn, line 107: I have no idea what I'm doing with this CSS. I would like to have the labels under the check boxes like we have in SW1.

SeanArchibald: There is a div element which wraps around the 16 span elements for the input channels checkboxes. If you give that an id such as input-channels-wrapper or something, then you can add this css:

```
#input-channels-wrapper {
    column-gap: 1rem;
}
#input-channels-wrapper span {
    display: flex;
    flex-flow: column;
    text-align: center;
}
```

I think that will make the check boxes appear the way you want.

Screenshot_20220702_142452

frostburn:
Fixed the channel checkbox layout.

frostburn:
Approved on Discord.

PR #108, opened 2022-06-25
Reaper exporter

frostburn:


src/exporters/__tests__/reaper.txt
```diff
@@ -0,0 +1,129 @@
...
+88 3700.000 (3)
+87 8/1 (3)
+86 3486784401/819200000 (2)
+85 5.437 (2)
+84 20/3 (2)
+83 14\5 (2)
+82 2500.000 (2)
+81 4/1 (2)
+80 3486784401/1638400000 (1)
+79 2.718 (1)
```
frostburn, line 50: This should use a decimal comma instead.

SeanArchibald (commented):
Is this missing the Reaper export button, or is that coming in a later change?

frostburn:
Modals didn't exists when I made this. I can do it here or in a later PR. Your call.

SeanArchibald:
Ah yes fair enough. This is approved then!

PR #120, opened 2022-06-27
Encode app state in URL

frostburn:
State decoding is registered as a navigation hook so that the browser back button acts as an undo.


frostburn:
Fixed the issue where the first encoding of the url fails.

SeanArchibald (approved):

frostburn:
I'll do a bit more self-review to check that the newly merged extended line types are encoded as intended.

frostburn:
Added special handling for the empty app state so that it can be undoed to. Self-approving that and merging.

Issue #137, opened 2022-07-02
Extend enumerate chord to all line types


frostburn:
SW1 supports decimals and other line types in chord enumeration.
### Changes needed to reach parity:
- [x] Rewrite `Scale.fromChord`
- [x] Add parsing support for bare numbers (but do not enable in scale builder)

Issue #115, opened 2022-06-26
Apply VS Code suggestions


frostburn:
VSCod(e/ium) is being a bit random with red underlining. Things are OK when making the pull requests, but later the editor decides that `setTimeout` doesn't return a `number` after all or the imports in tests need to be relative.

PR #142, opened 2022-07-03
Implement dedicated format-aware interval class

frostburn:
Remove unison line type
Remove unnecessary line types from importer results
Move reverse parsing to a dedicated class and adapt the rest of the application accordingly
Apply VSCodium suggestions
Extend enumerate chord to all line types
Implement scale repetition
Add support for multiple periods in rank 2 scale generation

refs https://github.com/xenharmonic-devs/scale-workshop/issues/115 https://github.com/xenharmonic-devs/scale-workshop/issues/137


frostburn:
Maybe `Interval`  is a better class name than `ScaleLine`.

src/scale-line.ts
```diff
@@ -0,0 +1,425 @@
...
+    options?: ScaleLineOptions
+  ) {
+    this.monzo = monzo;
+    this.lineType = lineType;
+    this.options = options || {};
+    this.name = name || this.toString();
+  }
+
+  zeroed() {
+    return new ScaleLine(this.monzo.mul(0), "unison", "1/1", this.options);
```
frostburn, line 59: This should preserve line type. (See above comment about unison.)

SeanArchibald (approved):

Issue #10, opened 2022-05-05
Reduce odd-limit approximation complexity


frostburn:
The algorithm for finding the closest odd-limit approximant is currently of complexity O(n²). It's unlikely that it's optimal.

PR #151, opened 2022-07-07
Reduce odd limit approximation complexity

frostburn:
Uses caching and floating point modular arithmetic to speed up approximating monzos by odd limit ratios

ref #10


frostburn:
Upon approval I will move the utility function to `xen-dev-utils`, but it makes sense to review the changes here.

frostburn:
This functionality was moved to `xen-dev-utils`. Closing this PR as this style of approximation relates non-destructive editing which we don't support yet.
The upcoming rational approximation modal requires an array of candidates instead of a single best one.

Issue #111, opened 2022-06-25
Implement missing modals for scale creation / modification


frostburn:
### Creation
- [x] Equal temperament
- [x] Subharmonic series segment
- [x] Enumerate chord
- [x] Combination product set
- [x] Lattice (cube)
- [x] Cross polytope
- [x] Euler genus
- [x] Dwarf scale
- [x] Octaplex (easter egg)

### Modification
- [x] Sort
- [x] Reduce
- [x] Rotate
- [x] Subset
- [x] Stretch
- [x] Random variance
- [x] Approximate by ratios
- [x] Approximate by subharmonics
- [x] Equalize

PR #161, opened 2022-07-10
Enumerate chord

frostburn:
Implement parsing of colon separated chords 
Implement parsing of bare numbers
Implement chord enumeration modal 

refs #111 #137


frostburn:
Now that CPS supports arbitrary factors we need a parser for that. I'll extend this chord parser to handle other separators.

SeanArchibald (approved):

PR #166, opened 2022-07-11
Do not allow zero to be converted to monzo

frostburn:
Prevent infinite loops by throwing an error for zero


frostburn:
Fixes the application getting stuck when accidentally pressing 0 while enumerating a chord.

SeanArchibald (approved):

PR #167, opened 2022-07-11
Implement a modal for spanning lattices

frostburn:
ref #111


frostburn:
Changed the labels to reflect existing language about generators up from 1/1 in other modals.

SeanArchibald (approved):

Issue #30, opened 2022-05-09
Synth


frostburn:
Probably `Tone.js` but we can make custom audio worklets if Tone isn't xen enough.

frostburn:
We have a basic synth now. I'll make a new ticket when a more advanced synth can be integrated.

PR #168, opened 2022-07-11
Placeholder synth and waveform visualizer

frostburn:
ref #30

### DIFF
--- a/src/views/SynthView.vue
+++ b/src/views/SynthView.vue
@@ -1,5 +1,11 @@
 <script setup lang="ts">
-defineProps<{
+import { computed, onMounted, onUnmounted, ref } from "vue";
+import TimeDomainVisualizer from "@/components/TimeDomainVisualizer.vue";
+
+const props = defineProps<{
+  audioContext: AudioContext;
+  audioOutput: AudioNode | null;
+  mainVolume: number;
   equaveShift: number;
   isomorphicHorizontal: number;
   isomorphicVertical: number;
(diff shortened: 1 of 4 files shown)


src/views/SynthView.vue
```diff
@@ -31,14 +59,45 @@ function updateEquaveShift(event: Event) {
...
+          <input
+            type="range"
+            min="0"
+            max="0.5"
+            step="any"
+            v-model="mainVolume"
+          />
+        </div>
+        <h2>Waveform</h2>
+        <TimeDomainVisualizer
```
frostburn, line 95: This canvas element is ugly here. How do you even style those responsively? I'm pretty sure it has to be a canvas due to the amount of sound data it has to render.

frostburn: Used CSS width = 100% with offset width compensation while drawing the canvas.

frostburn:
The order of the main lowpass and gain should be reversed to get visuals of the actual sound output.

frostburn:
**Sevish** _— Today at 4:37 PM_

> It's worth adding a high pass as well just in case someone tries playing 5 mHz and blows out their speakers 😄
> a steep one at 30 Hz would prob be fine

frostburn:
Added the highpass. Didn't set it too steep to avoid ringing. Could add another for 24db/octave if needed.

SeanArchibald:
The default main volume is loud enough that if I hold 5 notes then I hear distortion. Maybe it can come down to 0.11 ish? At this value I get roughly equal loudness with SW1 triangle.

If I change the value of main volume slider, and then change the value of any isomorphic key mapping control, the main volume slider moves by itself (though the actual gain doesn't seem to get affected).

frostburn:
> If I change the value of main volume slider, and then change the value of any isomorphic key mapping control, the main volume slider moves by itself (though the actual gain doesn't seem to get affected).

That seems like a bug in v-model I'll check if there's a workaround.

frostburn:
Set default volume to 0.11 and fixed the volume slider moving by itself.

frostburn:
Fixed the issues on Chromium.

SeanArchibald (approved):

Issue #75, opened 2022-06-20
Typing keyboard input and mapping


frostburn:
- [x] Isomorphic layout
- [x] White/black layout with accidentals
- [x] Octave +/- keys

Remember to make it configurable whether or not the key to the left of _Z_ exists.

PR #170, opened 2022-07-11
Implement piano style keyboard mapping constructed from key colors

frostburn:
Add relevant config to Synth tab
Add relevant URL encoding

refs #75 #182

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -19,6 +19,10 @@ import { debounce } from "@/utils";
 import { version } from "../package.json";
 import type { Interval } from "@/interval";
 import { arraysEqual } from "xen-dev-utils";
+import {
+  mapWhiteAsdfBlackQwerty,
+  mapWhiteQweZxcBlack123Asd,
+} from "./keyboard-mapping";
 
 // === Root props and audio ===
 const rootProps = defineProps<{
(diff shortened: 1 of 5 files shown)


src/App.vue
```diff
@@ -40,6 +44,9 @@ const keyColors = ref([
 ]);
 const isomorphicVertical = ref(5);
 const isomorphicHorizontal = ref(1);
+const keyboardMapping = reactive<Map<string, number>>(new Map());
+mapWhiteAsdfBlackQwerty(keyColors.value, keyboardMapping);
+const keyboardMode = ref<"isomorphic" | "mapped">("isomorphic");
```
frostburn, line 49: I intended "mapped" to be a free mapping, but #182 would warrant a third mode that only includes canonical layouts built from key colors and thus has a small URL footprint.

src/views/SynthView.vue
```diff
@@ -61,6 +71,8 @@ onUnmounted(() => {
     analyser.value = null;
   }
 });
+
+// TODO: Modal for custom mapping per keycode
```
frostburn, line 75: This should be done during this PR so that the URL scheme can be designed.

frostburn:
Implemented URL encoding and pushed TODO comments to #190 .
Ready for review again.

SeanArchibald (approved):

PR #179, opened 2022-07-13
Implement modal for approximating scale degrees with ratios

frostburn:
Update xen-dev-utils dependency
Update moment-of-symmetry (co)dependency
Remove odd limit approximation from monzo and scale as obsolete and unnecessary

refs #10 #111

### DIFF
--- a/src/components/modals/modification/ApproximateByRatios.vue
+++ b/src/components/modals/modification/ApproximateByRatios.vue
@@ -1,0 +1,292 @@
+<script setup lang="ts">
+import type Scale from "@/scale";
+import { computed, ref, watch } from "vue";
+import Modal from "@/components/ModalDialog.vue";
+import {
+  approximateOddLimitWithErrors,
+  approximatePrimeLimitWithErrors,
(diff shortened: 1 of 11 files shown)


frostburn:
Limit max cents to 600.
Always show at least one approximation. Set state to invalid if it's above max cents.

src/components/modals/modification/ApproximateByRatios.vue
```diff
@@ -0,0 +1,286 @@
...
+    return [{ fraction, error, limit: primeLimitString(fraction) }];
+  }
+  return approximationsAndErrors.map(([fraction, error]) => ({
+    fraction,
+    error,
+    limit: primeLimitString(fraction),
+  }));
+});
+
+watch(approximationsWithErrorsAndLimits, (newValue) => {
```
frostburn, line 130: These watchers seem to trigger even when the modal is not loaded. That's a problem in terms of performance...

SeanArchibald (approved):

Issue #184, opened 2022-07-16
Make it possible to use large subgroups for rank 2


frostburn:
While it makes sense not to dynamically calculate huge temperaments, they are now impossible to use as you cannot progress into the first "tab" of the modal without clicking on a dynamically generated MOS size.

Try 136/135, 169/168, 176/175, 190/189, 221/220, 256/255

PR #187, opened 2022-07-16
Add a button to force MOS size calculation with big subgroups

frostburn:
ref #184


frostburn:
Reviewed on Discord. :computer: :fire:

Issue #163, opened 2022-07-10
Convert modifier/modal


frostburn:
Convert intervals to plain cents or monzos.
Get rid of preferences to see intervals in lowest terms.

PR #188, opened 2022-07-17
Implement a modal for converting lines to other types

frostburn:
Add associated methods to Interval and Scale

ref #163

### DIFF
--- a/src/components/modals/modification/ConvertType.vue
+++ b/src/components/modals/modification/ConvertType.vue
@@ -1,0 +1,144 @@
+<script setup lang="ts">
+import type Scale from "@/scale";
+import { onMounted, ref } from "vue";
+import Modal from "@/components/ModalDialog.vue";
+import type { IntervalOptions, IntervalType } from "@/interval";
+import { Fraction } from "xen-dev-utils";
+
+const props = defineProps<{
+  scale: Scale;
+  centsFractionDigits: number;
+  decimalFractionDigits: number;
+}>();
+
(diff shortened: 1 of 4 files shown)


src/components/modals/modification/ConvertType.vue
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  emit("update:scale", props.scale.asType(type.value, options));
+}
+</script>
+
+<template>
+  <Modal @confirm="modify" @cancel="$emit('cancel')">
+    <template #header>
+      <h2>Convert line types</h2>
```
SeanArchibald, line 57: This is a useful new feature and works well on my local!

It might be the first time we have exposed users to the idea of "line types" as far as I remember. Are any these alternative phrases easier for average users to grasp? Convert interval types, convert interval notation, convert scale notation... Any other suggestions?

frostburn: The closest I've seen SW1 coming to naming these in user space is *value* in the User Guide.
> Values with a . are cents values: e.g. 701.955
> Values with slash (/) are ratios: e.g. 3/2
> Values with a backslash (n\m) are n degrees of m-EDO: e.g. 7\12
> Values with comma (,) are decimals: e.g. 1,5

So maybe *convert interval values*?

frostburn:
The modal is now called *Convert interval values* and the select element is labeled as *Interval type*.

SeanArchibald (approved):

Issue #195, opened 2022-07-18
Panic


frostburn:
Turn off synth and MIDI.
Release all sustained keys.

PR #196, opened 2022-07-18
Add a panic button

frostburn:
ref #195

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -459,6 +459,38 @@ function updateMidiInputChannels(newValue: Set<number>) {
   midiInputChannels.clear();
   newValue.forEach((channel) => midiInputChannels.add(channel));
 }
+
+function panic() {
+  console.log("Firing global key off.");
+  // Variable name typo fixed in #170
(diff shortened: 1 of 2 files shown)


SeanArchibald:
Not sure if this is expected behaviour but the panic button seems to cycle through 3 states. The button displays:

Quiet - can play notes as normal
Quiet! - any held notes get stopped. But you can start playing new notes straight away
Resume - can't play any notes

frostburn:
> Not sure if this is expected behaviour but the panic button seems to cycle through 3 states.
This is intended. The first press of the panic button cuts all notes, the second press cuts all sound and the third press returns everything back to normal.
Do you like it? We could make a hard panic button elsewhere if soft panic is enough for the main UI.

frostburn:
Stateless panic is enough, but hard panic can be left in the codebase in case it's needed one day.

frostburn:
Made panic stateless. Also added a missing MIDI to synth deactivation.

SeanArchibald (approved):

Issue #146, opened 2022-07-05
Basic Analysis


frostburn:
Add something to the Analysis tab.

- [x] Modes / Interval matrix
- [x] Animated chord wheel of the currently played chord

PR #200, opened 2022-07-19
Add content to Analysis tab

frostburn:
Add a table for the interval matrix of the scale
Add a chord wheel visualizer for the currently played chord

ref #146

### DIFF
--- a/src/views/AnalysisView.vue
+++ b/src/views/AnalysisView.vue
@@ -1,21 +1,205 @@
 <script setup lang="ts">
+import { intervalMatrix } from "@/analysis";
 import type Scale from "@/scale";
+import type { VirtualSynth } from "@/virtual-synth";
+import ChordWheel from "@/components/ChordWheel.vue";
+import { computed, ref } from "vue";
+import type { Interval, IntervalOptions } from "@/interval";
 
-// Placeholder code for analysis to see that props are passed to lazy views too.
-defineProps<{
-  scaleLines: string[];
+const props = defineProps<{
   scale: Scale;
+  virtualSynth: VirtualSynth;
 }>();
+
+const cellFormat = ref<"best" | "cents" | "decimal">("best");
+const trailLongevity = ref(70);
+const maxOtonalRoot = ref(16);
+const maxUtonalRoot = ref(23);
+
+const fadeAlpha = computed(() => 1 - trailLongevity.value / 100);
+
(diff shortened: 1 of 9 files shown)


src/views/AnalysisView.vue
```diff
@@ -1,21 +1,120 @@
...
+          <tr v-for="(row, i) of matrix" :key="i">
+            <th>{{ scale.getInterval(i).name }}</th>
+            <td v-for="(name, j) of row" :key="j">{{ name }}</td>
+          </tr>
+        </table>
+        <h2>Currently played chord</h2>
+        <div class="minor-columns-container">
+          <div class="minor-column wheel-container">
+            <h3>Otonal</h3>
+            <div class="control-group-is-broken-here-for-some-reason">
```
frostburn, line 43: I have no idea why the nested columns are not working. Little help here?

frostburn:
Removed the nested column layout. Leaving design to the UI pass.

SeanArchibald:
Chord wheel graphics are awesome of course!

The interval matrix seems to work well for a lot of scales, but I did find some weird edge cases.

If the current scale is an equal temperament e.g. `1\5 2\5 3\5 4\5 5\5` then the first degree appears as `0\1<1>` which looks difficult to read. Maybe unisons could appear as 0\1 or even 1/1? Apparently Scala doesn't even display the unisons in its interval matrix. That could be an option too.

Should counting start from 0 or 1? Looks like currently the unison is counted as the 1st but in my head it makes better sense as the 0th because it's 0 degrees up from the unison.

When the scale has composite intervals of ratios + cents, some of the results seem wrong.

```
9/8
5/4
4/3
3/2
9/8 + 700.
5/4 + 700.
4/3 + 700.
```

Likewise for scales of mixed interval types.

```
9/8
400.
9\22
[-1, 1>
27/16
7/4
1200.
```

I can see this getting complicated so one option for these scales is convert intervals to cents.

frostburn:
The `0\1<1>` issue has been fixed in `main`.
I'll look into displaying cents better.

frostburn:
Added more sanity to the matrix cell formatting, but the intended result is only visible with #198 . Good for review after merging that.

frostburn:
The matrix should now appear as intended.
Replaced compute-intensive rolling hard fade out with an intermittent strong blend.

SeanArchibald (approved):

Issue #135, opened 2022-06-30
Generate rank 2 - flip generator button


frostburn:
The ability to flip a generator from bright to dark.
Show if generator is bright.
Support multiple periods per equave.

frostburn:
Requires an update to `xen-dev-utils` so maybe merge #179 first to avoid conflicts.

PR #215, opened 2022-07-29
Add buttons to flip generators

frostburn:
ref #135


frostburn:
There are some issues with the layout and dependency versioning. Will fix after the UI pass.

frostburn:
The button in rank 2 is a bit jumpy, so I'm probably missing something about the styling.

frostburn:
Switching branches...

Issue #136, opened 2022-06-30
Assing numpad +- to degree shift


frostburn:
Same as * and / except a movement of 1.

frostburn:
These could be added to the URL so it would be preferable to merge #170 first.

PR #217, opened 2022-07-30
Assing numpad +- to degree shift

frostburn:
ref #136


SeanArchibald (approved):
Works as expected, very useful feature

PR #224, opened 2022-07-31
UI refinement

SeanArchibald:


src/assets/base.css
```diff
@@ -246,8 +252,22 @@ ul.btn-group, .btn-dropdown-group ul {
...
+}
+.section:hover::after {
+  color: var(--color-accent-text);
+  background-color: var(--color-accent);
+  border-color: var(--color-accent);
 }
+p.section {
+  font-weight: bold;
+  cursor: pointer;
+}
```
frostburn, line 273: Missing newline at the end of the file.

src/components/ScaleBuilder.vue
```diff
@@ -349,6 +342,18 @@ async function doImport(importerKey: ImporterKey, event: Event) {
           />
           <span>A5</span>
         </div>
+
+        <!--<div class="control">
```
frostburn, line 346: Commented out code should state the reason for it's existence or be removed completely.

src/components/ScaleBuilder.vue
```diff
@@ -450,7 +466,7 @@ async function doImport(importerKey: ImporterKey, event: Event) {
         "
       >
         <p><strong>Share scale as URL</strong></p>
-        <p>Open the sharing dialog</p>
+        <p>Copy this scale's unique URL for convenient sharing</p>
```
frostburn, line 469: I think URL sharing has higher priority and should be moved up in the list. Now I have to scroll to share my scale if I don't feel like copying from the address bar.

frostburn:
Can you take a look at the MOS dialog's EDO section? It jumps up and down depending on the number of scales found resulting in janky UX. (Or make a new issue if that's out of the scope of this PR.)

src/components/modals/modification/RotateScale.vue
```diff
@@ -18,9 +18,9 @@ function modify() {
       <h2>Rotate</h2>
     </template>
     <template #body>
-      <p>Rotates the mode of your scale.</p>
-      <p>The resulting scale will be sort ascendingly.</p>
       <div class="control-group">
+        <p>Rotates the mode of your scale.</p>
+        <p>The resulting scale will be sorted ascendingly.</p>
```
frostburn, line 23: I copied this over from SW1 and I thinks it's the first time I've seen the word *ascendingly*. I'd rather replace it with *in ascending order*.

src/views/AboutView.vue
```diff
@@ -8,35 +8,45 @@
...
-    <ul>
-      <li>
-        <a href="https://github.com/xenharmonic-devs/scale-workshop">GitHub</a>
-      </li>
-      <li>
-        <a href="https://github.com/xenharmonic-devs/scale-workshop/issues"
+    <div class="control-group">
+      <div class="control">
+        <p>
+          With contributions from Sevish, Lumi Pakkanen, Lajos Mészáros,
```
frostburn, line 18: The last comma should be replaced with *and*.

src/views/AboutView.vue
```diff
@@ -8,35 +8,45 @@
...
+        </p>
+      </div>
+    </div>
+    <div class="control-group">
+      <div class="control">
+        <a class="btn" href="https://github.com/xenharmonic-devs/scale-workshop">GitHub</a>
+        <a
+          class="btn"
+          href="https://github.com/xenharmonic-devs/scale-workshop/issues"
           >Report issue</a
```
frostburn, line 29: *Report an issue* would be more grammatically correct, but might look verbose on a button.

src/views/AnalysisView.vue
```diff
@@ -156,50 +179,66 @@ const matrix = computed(() => {
...
+}
+
+/* Chord wheel */
+.chord-wheel {
+  border: 1px solid var(--color-border);
+  max-width: 100%;
+  height: auto;
+}
+[data-theme="dark"] .chord-wheel {
+  filter: invert(1);
```
frostburn, line 242: While nifty, I would prefer using the correct colors in the canvas itself. Even dark mode can use red to indicate otonality.

frostburn:
There's some code churn e.g. with the dark theme that should be squashed away.

frostburn:
The dark theme doesn't seem to connect with the toggle in User preferences. Add a TODO or file an issue (preferred).

src/views/SynthView.vue
```diff
@@ -195,7 +211,11 @@ div.keyboard-controls {
 .waveform {
   width: 100%;
   height: auto;
+  border: 1px solid var(--color-border);
+  border-radius: 5px;
 }
 
-/* TODO: media queries left for a later UI pass */
+[data-theme="dark"] .waveform {
+  filter: invert(1);
```
frostburn, line 219: Too nifty for my taste. I'd rather see colors passed as component properties.

SeanArchibald:
> Can you take a look at the MOS dialog's EDO section? It jumps up and down depending on the number of scales found resulting in janky UX. (Or make a new issue if that's out of the scope of this PR.)

The current design for the modals is that they adapt size to contain the content in them. We could force the size of this particular modal so that it behaves differently from the others. I'm not sure I really have much of an issue with the modal changing size so not sure what to do with this one yet.

SeanArchibald:
> The dark theme doesn't seem to connect with the toggle in User preferences. Add a TODO or file an issue (preferred).

Is #226 the fix?

src/components/ScaleBuilder.vue
SeanArchibald: Have added comment

src/views/AnalysisView.vue
SeanArchibald: I've removed this filter for now. Once the dark/light/default switch behavior is merged in, I think that will be a good time to come back to this

frostburn:
> Is #226 the fix?
Yes. Either merge the pull request, cherry pick from the commit or just copy & paste the code to apply the fix.

frostburn:
> The current design for the modals is that they adapt size to contain the content in them. We could force the size of this particular modal so that it behaves differently from the others. I'm not sure I really have much of an issue with the modal changing size so not sure what to do with this one yet.

The issue is that I cannot keep clicking on the up/down arrows on the number input because the input element keeps moving under the mouse pointer.
Fixing the modal size sounds like a good solution. It only needs to fix the size of this particular "tab" so maybe a transparent element could do the trick to force maximum size there.

SeanArchibald:
> 

I submitted issue #231 for the modal resize problem and will come back to it after thinking through some potential solutions. Thanks for all the feedback!

Think everything else is ok for reviewing again. There will be more UI things for me to do shortly (and before SW2 release) but I'd quite like to get all these changes safely into master. Maybe we can have another call some time in future to figure out the next steps

frostburn:
I have no further comments on looks, function or code style. Run the linter and clean the commit history using rebase/squash. Should be good to go after that. :+1: 
A call sounds like a good idea.

frostburn:
This PR has been obsoleted.

PR #237, opened 2022-08-07
Improve the temporary synth

frostburn:
ref #30


frostburn:
A bit different from SW1 and doesn't interact with the URL, but good enough for beta testing.

SeanArchibald (approved):

Issue #233, opened 2022-08-07
Add a link to the wiki for the user guide tab


frostburn:

frostburn:
Replace user guide tab with just a button/link on the first tab.

PR #242, opened 2022-08-10
Replace user guide tab with a link to the wiki in About view

frostburn:
ref #233

### DIFF
--- a/src/views/AboutView.vue
+++ b/src/views/AboutView.vue
@@ -22,6 +22,12 @@
     </div>
     <div class="control-group">
       <div class="control">
+        <a
+          class="btn"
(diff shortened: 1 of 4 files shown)


src/views/AboutView.vue
```diff
@@ -22,6 +22,11 @@
     </div>
     <div class="control-group">
       <div class="control">
+        <a
+          class="btn"
+          href="https://github.com/xenharmonic-devs/scale-workshop/wiki"
```
SeanArchibald, line 27: I would suggest adding `target="_blank"` to open the guide in a new tab. After that, looks good to merge in

Issue #185, opened 2022-07-16
Make sure that modal's computed properties only run when the modal is active


frostburn:
With a large amount of modals this is a performance issue.

frostburn:
It seems that watchers are to blame. The watched properties will need to come with guards if they're expensive to calculate.

frostburn:
Make a util for modal watchers.

PR #245, opened 2022-08-11
Only compute modal properties when modals are active

frostburn:
ref #185


frostburn:
There are a few more watched properties like these, but they are even cheaper to compute. It's arguable if even these need guards. The worst offender was already taken care of previously.

SeanArchibald (approved):

PR #247, opened 2022-08-13
Add buttons to flip generators

frostburn:
Tweak the control group hierarchy for rank-2 modal

ref #135


frostburn:
Fixed the issue where flipping would eat generators in cents. Now it does add an extra `1/1 + cents` after flipping back, but should work just fine.

SeanArchibald (approved):

PR #249, opened 2022-08-14
Fix incorrect placement of input field and buttons within div.control

SeanArchibald:


frostburn:
Run the linter as suggested by the CI. Good to merge after that. :+1:

PR #258, opened 2022-08-19
Make MOS generator pyramid responsive for small screens

SeanArchibald:


frostburn:
Could you also make the pyramid buttons of equal size in the small layout. Now buttons with single vs. double digits don't align.
Also can you add a line to the commit message saying ref #236 for future historians?

SeanArchibald:
Pyramid buttons are now equal size on small layout, and I have amended the commit message to mention #236

frostburn (approved):

Issue #266, opened 2022-08-25
Add synth envelope to URL


frostburn:
Seems likely that we ship with the current synth.
Parse synth envelopes from SW1 URLs.
URL-encode envelope params.

PR #267, opened 2022-08-25
Encode synth state in URL

frostburn:
Parse synth state from SW1 URLs

ref #266


frostburn:
Needs to support the new waveforms.

SeanArchibald (approved):

Issue #262, opened 2022-08-22
Update TAMNAMS names with current wiki content


frostburn:
Also implement naming extensions for at least 11-note and 12-note scales based on my (Lumi's) proposal.
(Handled by the `moment-of-symmetry` package.)

PR #273, opened 2022-08-26
Use new TAMNAMS names and extensions

frostburn:
Replace subset based system with anti-pairs and family tree names

refs #262


frostburn:
The MOS pyramid should now use the new names for scales <= 10 notes and extended names like *m-chro machinoid* for larger scales such as 6L 5s.

SeanArchibald (approved):

Issue #261, opened 2022-08-22
Tweak wart syntax


frostburn:
Make first letters of the alphabet correspond to primes. Make letters from q onward correspond to strict formal primes.
(This is handled in the `temperaments` package.)

PR #274, opened 2022-08-27
Tweak wart syntax

frostburn:
Make first letters of the alphabet correspond to primes.
Make letters from q onward correspond to strict formal primes.

ref #261


frostburn:
It's about this stuff: https://en.xen.wiki/w/Val#Shorthand_notation

SeanArchibald (approved):

PR #294, opened 2022-09-13
Implement finite synth polyphony

frostburn:
ref #290

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -69,7 +69,6 @@ const equaveShift = ref(0);
 const degreeShift = ref(0);
 const heldNotes = reactive(new Map<number, number>());
 const typingActive = ref(true);
-const synth = reactive(new Synth(rootProps.audioContext));
 const midiInput = ref<Input | null>(null);
 const midiOutput = ref<Output | null>(null);
 const midiInputChannels = reactive(new Set([1]));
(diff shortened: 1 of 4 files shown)


frostburn:
Breaks when deployed. Need to debug.

frostburn:
Fixed issues related to initialization on fist user gesture. Having the oscillators always on seems to fix #289 as well.
Good for review again.

SeanArchibald:
Looking good for the most part. Got some unexpected behaviour in testing. I set the max poly to 2 to make it really easy to hear and trigger the voice stealing behaviour. Something odd happens when using the shift key to sustain notes. After I hit the 3rd note, the 1st voice is stolen as expected, but if I try to retrigger the 1st note then it doesn't play. If I try to then retrigger the 2nd note then all voices are silent.

I suspect that sustained notes which have had their voices stolen are not being updated, but haven't dug into the code yet.

frostburn:
Voice stealing (`Synth` class) doesn't communicate back to the `Keyboard` class. It still thinks notes are playing even if their voice gets stolen. A second press of a sustained note simply turns it off, so you'd need multiple key presses to re-trigger. I'll see if there's an easy solution that more closely matches the sound that is produced.

frostburn:
With max polyphony set to 1 and playing two notes there is a clear bug where releasing the first note causes the second one to stop playing.

frostburn:
Fixed the issue with note offs of stolen voices. I don't think the sustain issue is a bug. The capabilities of the Web Audio synth shouldn't limit MIDI output so the current configuration where `Synth` doesn't talk to `Keyboard` is the correct one IMO.

SeanArchibald (approved):

PR #297, opened 2022-10-05
Add status indicators tray to top bar

SeanArchibald:


frostburn:
The indicator is nice, but it's not ideal from a troubleshooting point of view. If the focus is on a select element and the user is wondering why there's no sound they're left in the dark unless they saw the indicator before and remember what it said.
The indicator is less useful on mobile as most smart phones don't have a physical keyboard. I would hide it on small screens.

frostburn:
Requested changes deferred to #299

Issue #144, opened 2022-07-04
Add sanity checks to user input


frostburn:
In addition to reasonable `min` and `max` values on input elements, add sanity checks to scale generation code so that the application doesn't hang.

# Checklist
### Combination product set
The user would need to enter thousands of factors to cause performance issues. No need to fix.
### Cross Polytope
Same as CPS.
### Dwarf Scale
It's too easy to enter huge numbers. Needs a limit.
### Enumerate Chord
No limits needed.
### Equal Temperament
Number of automatically generated notes needs to be limited.
### Euler Genus
Huge guide tones cause issues.
### Harmonic Series
Needs a limit on the number of notes.
### MOS Scale
Needs limit and/or rounding on everything.
### Rank 2
Might need some limits somewhere, but it's bit too complex for now. Make a new issue if needed in the future.
### Span Lattice
Same as CPS
### Subharmonic Series
Same as Harmonic Series
### Octaplex
No limits needed.

### Approx Harmonics
No limits needed.
### Approx Ratios
Odd and prime limits need caps.

PR #306, opened 2022-10-13
Add sanity checks for user input

frostburn:
ref #144

### DIFF
--- a/src/components/modals/generation/DwarfScale.vue
+++ b/src/components/modals/generation/DwarfScale.vue
@@ -3,16 +3,18 @@ import { DEFAULT_NUMBER_OF_COMPONENTS } from "@/constants";
 import Scale from "@/scale";
 import { ref } from "vue";
 import Modal from "@/components/ModalDialog.vue";
+import { clamp } from "xen-dev-utils";
 const emit = defineEmits(["update:scale", "update:scaleName", "cancel"]);
 const val = ref(12);
 const equave = ref(2);
 function generate() {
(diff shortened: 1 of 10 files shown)


SeanArchibald (changes requested):
Nice stuff, I didn't realise until now but this is a really good improvement to have.

I tried to create 9999ed2. What I ended up with was the clamped 1024ed2 which is fine but the scale name was still set to "9999 equal divisions of 2/1". The HTML input field itself could have max="1024" too.

Trying to create harmonics 1-10001 still takes a couple minutes on this 2015 MacBook Air, then page performance continues to be really bad. I'm wondering if the limit should be more strict. Even 1-1000 was very slow.

For approximate by ratios > odd limit, I tried to type in 300 but the page instantly froze with 30 in the input box. It did come back after a couple mins. To replicate the issue again a page refresh is needed. On my phone, I had a much larger scale active and had to force close the tab.

frostburn:
The ET modal is a difficult beast to tame. A subset of 9999ed2 is fine so we can't have a hard `max="1024`. I'll check if anything can be done about the scale name.

frostburn:
I was only testing on a beefy desktop computer. I'll try tuning the limits to be more mobile friendly.

frostburn:
Fixed the issues. Deploying and testing on a smartphone...

frostburn:
My phone didn't blow up. Good for review again.
I'm not worried about large performance drops. The user is shooting themselves in the foot when running into the limits anyway.

Issue #329, opened 2022-11-02
Notes stay pressed visually


frostburn:
**Carmen (Bragtime)** — Today at 2:59 PM

> [...] sometimes when I press a note down and unpress it the note visually stays colored in as if I still had it pressed just thought I'd let you know 🤔

PR #330, opened 2022-11-02
Make tuning table highlight hacks support tabs

frostburn:
Store hack state outside of hot reloading
Mount rows as active if keys pressed when switching to main tab

ref #329


forrcaho (approved):
Reproduced the problem and confirmed it was fixed. I only have a vague idea how this code works, but I didn't see anything questionable.

frostburn:
Definitely one of the hairiest bits of SW. It might be possible to avoid these kind of hacks with better understanding of Vue.

Issue #339, opened 2022-11-16
import .scl not working when string after pitch value exists


kingemir:
https://huygens-fokker.org/scala/scl_format.html states "Anything after a valid pitch value should be ignored." (Ignored for setting up the scale itself). In practical use, people use this rule to put the name of the pitch after the pitch value. In the specifications website of the scl format (see link above), some valid examples are mentioned: 
100.0 C#
 5/4   E\

Some software even import these strings into a pitch name column, for example Lingot. 

Unfortunately, the sevish scale-workshop rejects scl files containing strings after the pitch value during import, even though these files are valid according to the format specifications.

frostburn:
Thank you for bringing this to our attention. We'll fix Scala import in the next release.

@kingemir If you want to make sure that your .scl files are imported correctly you should attach some here so that we can use them for testing.

kingemir:
[scl_files.zip](https://github.com/xenharmonic-devs/scale-workshop/files/10035732/scl_files.zip)
for testing

PR #342, opened 2022-11-17
Make the Scala importer more compliant with the spec

frostburn:
ref #339


src/parser.ts
```diff
@@ -365,6 +367,9 @@ export function parseLine(
     case LINE_TYPE.COMPOSITE:
       return parseComposite(input, numberOfComponents, options);
     default:
+      if (admitBareNumbers && lineType === LINE_TYPE.NUMBER) {
```
forrcaho, line 370: Instead of having `LINE_TYPE.NUMBER` handled with an if statement in the default section of the switch statement here, it might be more consistent to have a case for `LINE_TYPE.NUMBER` just like the other types. The catch is we'd have to handle the case where `admitBareNumbers` is false somehow: that could fall thru to the default case to throw the exception, or (if fall thru is considered unintuitive) there could be a throw in both sections.
 
 TBH, I'd rather have a dispatch table than a switch statement here (although maybe that's just me). What we have here definitely works, and I'm tempted to accept it and just make a note we'd like to refactor this in the future.

forrcaho (approved):
I had one comment about the switch statement in parser.ts, but what is there works and may be refactored to a dispatch table anyway, I think we can just keep this code for now.

src/parser.ts
frostburn: The parser has been adapted from Scale Workshop 1.5. I think switch statements should be generally avoided, but such refactoring is outside the scope of this PR. I'll make an issue in `scale-workshop-core` as this stuff is migrating there now.

Issue #287, opened 2022-09-01
Replace top export with simple URL copy to clipboard


frostburn:
Move the modal with email etc. to the bottom.

**Sevish** — Today at 6:10 PM

> random thought on share scale URL button
> we show a modal because it allows us to show buttons for email, twitter, facebook
> I wouldn't be surprised if these are never used
> I wonder if the button should simply copy the URL straight away
> then it would be super convenient for the usual XA crowd

frostburn:
**Sevish** — Today at 8:49 PM

> I clicked a copy button on Spotify yesterday and it updated the button text to say copied
> That would be an easy solution, but I guess it depends on what was in the button before

PR #346, opened 2022-11-20
Replace top export with simple URL copy to clipboard

frostburn:
ref #287


forrcaho (approved):
This looks fine to me (although I wonder if the modal dialog is actually needed at all).

frostburn:
The dialog is used for email and social sharing. We don't track our users so it's not possible to say if it's actually needed. Merging this change.

Issue #349, opened 2022-11-24
Note gets stuck if key released outside of virtual keyboard


frostburn:
**To replicate:**
1. Go to *Virtual Keyboard* tab.
2. Press and hold mouse button inside virtual keyboard.
3. Move mouse cursor to the top menu.
4. Release mouse button.
5. Press and release mouse button inside virtual keyboard.

**Result:** Note keeps playing. Even pressing *Panic* doesn't stop it.
**Expected result:** Note stops when moving outside of virtual keyboard with no residual stuck behavior. Pressing *Panic* stops all sound.

PR #355, opened 2022-11-25
Fix stuck notes

frostburn:
Keep track of synth note off callbacks for panic
Make virtual keys react to all mouse up events

ref #349


forrcaho (approved):
Sure, this looks good. I've confirmed it fixed the bug for me locally, and the code change makes sense (although tbh I'm still working on getting up to speed on event handler stuff).

Issue #345, opened 2022-11-20
Computer keyboard synth doesn't play after adjusting any slider on 'Synth' tab


forrcaho:
When I first bring up scale-workshop in my browser (Firefox on Win 11), my computer keyboard keys play notes. (By default, they're all 440 Hz, but that's not the issue here). I switch to the `Synth` tab in the top menu, and they still play notes.

If I adjust any of the sliders on the `Synth` page, my keyboard keys no longer create notes. When I switch to any other menu tab (e.g. `Virtual Keyboard`) the notes from my keyboard will start working again, with the new settings. I can also switch from `Synth` to any other tab and back before I try my keyboard again, and this also fixes the issue.

Only touching the sliders seem to reliably produce the error. There is similar behavior when I change the waveform, but it seems to play a little bit before stopping (which can then be fixed in the same way, by switching tabs). Changing the other non-slider settings seem to cause the problems sometimes, but not reliably.

I have this problem both with code from `main` running locally, and the online versions at https://sevish.com/scaleworkshop/synth?version=2.0.0 and https://scaleworkshop.lumipakkanen.com/ .

I also checked Chrome on Win 11, and it has the same issue.

forrcaho:
I found the `Key` text on the right side of the menu bar, and found out I can click anywhere on the `Synth` page to get the key activity back. This is "working" but

- it's not very discoverable ( I only found it after reading the source code), and
- it still seems like that extra click shouldn't be needed. Maybe a timer could be set and the keys reactivated when there are no settings changes for 1000 milliseconds or so?

frostburn:
Slider are a new addition to version 2.0. You're correct that they don't need to prevent synth input. Only text inputs and drop-down select elements need that because they have browser actions associated with key presses.

frostburn:
Discoverability is an issue, but at least the behavior is documented: https://github.com/xenharmonic-devs/scale-workshop/wiki/Synthesizer#playing-with-qwerty
Any ideas how to improve the situation?

forrcaho:
I commented on your slider fix PR; looks great, but we could also handle the radio buttons in the same line.  Either way, I think that could go into the code. 

I have some ideas about the other inputs that may or may not work:
* For `select` we could not capture any keyboard events, only mouse events. (I've experienced some weird behavior while focus was on the waveform select box where holding down a key would change the waveform and then start playing a note, but I can't reproduce it all the time. This would fix that issue.)            
[...]

PR #358, opened 2022-11-27
Allow users to play the synth while manipulating slider inputs

frostburn:
ref #345

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -521,6 +521,15 @@ function keyboardNoteOn(index: number) {
 // === Typing keyboard state ===
 function windowKeydownOrUp(event: KeyboardEvent | MouseEvent) {
   const target = event.target;
+  // Keep typing activated while adjusting sliders
+  if (
+    target instanceof HTMLInputElement &&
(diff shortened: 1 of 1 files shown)


src/App.vue
```diff
@@ -521,6 +521,12 @@ function keyboardNoteOn(index: number) {
 // === Typing keyboard state ===
 function windowKeydownOrUp(event: KeyboardEvent | MouseEvent) {
   const target = event.target;
+  // Keep typing activated while adjusting sliders
+  if (target instanceof HTMLInputElement && target.type === "range") {
```
forrcaho, line 525: We do have a radio button group as well, so it might make sense to make this 
``` 
if (target instanceof HTMLInputElement && ["range", "radio"].includes(target.type)) {
``` 

I also have an idea for the other inputs, which I'll put on the associated issue. But I think this could go in now, either as you have it or with my suggested modification.

forrcaho (approved):
Left a comment suggesting a possible change, but it's fine as it is also.

src/App.vue
frostburn: Good catch!

Issue #331, opened 2022-11-04
Optimize tempering for single vals (Modify scale)


frostburn:
It should be possible to support very high prime limits without much computational cost.
The default limit for single vals should be the maximum supported by `PRIMES`.

frostburn:
The ability to display the intervals as n-of-edo would be nice.
Maybe n-of-edji based on the current equave could be possible too.

PR #359, opened 2022-11-27
Implement high prime limit tempering for single vals

frostburn:
ref #331


frostburn:
To test load a JI preset and apply the *Temper* modifier with **Convert to edo-steps** enabled in the **Vals** radio option. You should end up with something that sounds like the original scale, but in the edo/val you specified and spelled using backslashes. This is similar to *Equalize* modifier but RTT aware.

frostburn:
Self-reviewing: Seems legit enough to merge. Will be replaced with a more consistent system in SW3 anyway.

Waiting for more important PRs to merge first so as not to cause any unnecessary rebasing.

Issue #326, opened 2022-10-31
Creating non-integer EDOs


MaxSamo:
Before the last update, it was possible to create non-integer EDOs, such as 52.99683 EDO instead of 53 EDO, 18.94809 EDO instead of 19 EDO, etc.
For context, 18.94809(...) EDO is very slightly more optimal tuning than 19 EDO, according to Riemann-Zeta tuning method

frostburn:
I think of such tunings as having a stretched octave/equave.

**Workarounds:**
- Create an equal temperament with the equave pre-stretched (replace 2/1 with 1202.3 etc.).
- Use the stretch/compress modifier after creating the scale.

**Feature proposal:**
Disable the *relative steps* and *absolute degrees* boxes when the user tries to create non-integer EDO, but let the generation go through.

MaxSamo:
I will also have to clarify that it was possible to create such tunings before the last update, so it should not take a lot to re-add this feature

frostburn:
Unfortunately the whole application was re-written from scratch so implementing this incidental feature will be pushed to version 2.1.0.
Meanwhile you can use the old version if non-integer EDOs are critical to your workflow: https://sevish.com/scaleworkshop1/

frostburn:
I think it's ambiguous how many notes a non-integer EDO scale should have. Version 1 breaks when you try to divide the octave into 0.5 parts even though the step-size is well-defined.
I suggest that non-integer scales are implemented as a single step only. They should sound the same as version 1 for divisions >= 1.

frostburn:
Core functionality done... https://github.com/xenharmonic-devs/scale-workshop-core
... but let's go with the single step approach.

MaxSamo:
Single step approach would surely work, as long as it's not hard to set the non-integer size of an EDO. Thanks for taking time on this!

PR #365, opened 2022-11-28
Make it possible to create non-integer equal temperaments as cETs

frostburn:
ref #326

### DIFF
--- a/src/components/ScaleBuilder.vue
+++ b/src/components/ScaleBuilder.vue
@@ -579,6 +579,8 @@ function copyToClipboard() {
 
     <EqualTemperamentModal
       :show="showEqualTemperamentModal"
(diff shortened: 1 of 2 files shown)


SeanArchibald:
The functionality works as expected for me, nice one!

I haven't marked this as approved yet because I agree the user experience needs a tweak. Here's my take on it...
When the user enters a non-integer value, I think the field should not show in red, because non-integer values are supported.
When a non-integer value is entered, the Relative steps and Absolute degrees fields should become greyed out, because they can only work for integer values.

frostburn:
Made the suggested changes. Also allowed cETs to be made using divisions less than 1.

frostburn:
Self-reviewing: Seems legit.

PR #369, opened 2022-12-02
Fix issues with multiple typing keyboard keyup callbacks

frostburn:


forrcaho (approved):
I was a little confused as to why this would be necessary; it looks like this is to handle multiple keyup events for one keydown? But the code changes are consistent and look like what you would do to handle that. Approved.

frostburn:
I was using this for the on-screen QWERTY keyboard I used in my YouTube tutorials and discovered the bug. It's not multiple keyups per keydown. It's multiple callbacks for keydown/keyup pairs. The registered keyups were not being called back correctly.

PR #381, opened 2022-12-10
Implement on-screen QWERTY keyboard

frostburn:


frostburn:
In response to seeing this in action in a tutorial video.

**@romeolz1948** on [YouTube](https://www.youtube.com/watch?v=U_yY5vwmk0I&lc=UgwNO2G9m965gNBaiuZ4AaABAg)

> Hey! Can you enable that 'YouTube' tab for the rest of us, that would be really handy...

SeanArchibald:
It works but I'm wondering if it would be easy for unmapped keys to be greyed out in the same way that the shift keys are greyed out. Currently unmapped keys show as white which is same as mapped keys
http://localhost:3000/scaleworkshop/qwerty?n=22edo%20orwell%5B9%5D&l=2Bm_5Bm_7Bm_aBm_cBm_fBm_hBm_kBm_mBm&c=~-~-~-~-~&k=p&p=1&a=2&y=3h&s=0&r=2s&version=2.0.1

frostburn:
A reasonable suggestion. I'll look into it.

frostburn:
Unmapped keys are already visually indicated in Piano mode. The shift keys are gray due to functionality not mapping. Are referring to non-functional keys that can never make sound in the application like *Tab*, *Caps Lock* or *Backspace*?

SeanArchibald:
Screenshot_20230325_195809

Screenshot attached shows the white keys that I had in mind. There is nothing visually to show that these unmapped keys are any different to the playable white keys on the row below. That confused me briefly during testing so i thought I would mention it. Otherwise all works as expected. So if you think it's not worth the change then I am happy for this to be merged in.

frostburn:
I see you're using the dark mode. I'm using light mode and everything is correctly indicated there. I simply forgot to check that the indicators work in dark too.

frostburn:
The indicators were in fact working as intended in dark mode, it's just that the human visual system wasn't picking up low contrast indicators against a high contrast background. Should be better now.

Issue #304, opened 2022-10-11
Optimize computed frequencies


frostburn:
While computing 128 unique frequencies isn't a huge performance issues it's still up there considering how often the state is updated while editing the scale.

Frequency calculation can be optimized by computing everything within a single equave and multiplying the result over to other frequency ranges. This eliminates costly monzo to frequency conversions.

frostburn:
Core functionality done: https://github.com/xenharmonic-devs/scale-workshop-core

PR #385, opened 2022-12-16
Optimize computed frequencies

frostburn:
ref #304


frostburn:
Relevant line in the core package: https://github.com/xenharmonic-devs/scale-workshop-core/blob/main/src/scale.ts#L606

SeanArchibald (approved):

Issue #239, opened 2022-08-10
Duplicate-at-offset modifier


frostburn:
Repeat the scale e.g. a third above with options on how to deal with intervals exceeding the equave.

frostburn:
Also look into filtering out intervals that exceed the equave in case that's useful for MV3 or something.

frostburn:
Core functionality done: https://github.com/xenharmonic-devs/scale-workshop-core

PR #386, opened 2022-12-16
Add a modal for merging an offset copy with the scale

frostburn:
ref #239

### DIFF
--- a/src/components/ScaleBuilder.vue
+++ b/src/components/ScaleBuilder.vue
@@ -28,6 +28,7 @@ import RandomVarianceModal from "@/components/modals/modification/RandomVariance
 import TemperModal from "@/components/modals/modification/TemperScale.vue";
 import EqualizeModal from "@/components/modals/modification/EqualizeScale.vue";
 import ConvertModal from "@/components/modals/modification/ConvertType.vue";
+import OffsetModal from "@/components/modals/modification/MergeOffset.vue";
 import ApproximateBySubharmonicsModal from "@/components/modals/modification/ApproximateBySubharmonics.vue";
 import ApproximateByRatiosModal from "@/components/modals/modification/ApproximateByRatios.vue";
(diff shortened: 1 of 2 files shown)


frostburn:
Relevant code in the core package: https://github.com/xenharmonic-devs/scale-workshop-core/blob/main/src/scale.ts#L902

SeanArchibald (commented):
Started with 5ed2
Merge offset > `400.` > Keep
http://localhost:3000/scaleworkshop/?n=5%20equal%20divisions%20of%202%2F1&l=1B5_B5SPSb4._2B5_1B5SPSb4._3B5_2B5SPSb4._4B5_3B5SPSb4._4B5SPSb4._5B5&c=~-~-~&f=5l&p=1&a=5&y=334&s=0&r=rs&version=2.0.1
Shouldn't `4\5 + 400.` come after `5\5`?

frostburn:
> Started with 5ed2 Merge offset > `400.` > Keep http://localhost:3000/scaleworkshop/?n=5%20equal%20divisions%20of%202%2F1&l=1B5_B5SPSb4._2B5_1B5SPSb4._3B5_2B5SPSb4._4B5_3B5SPSb4._4B5SPSb4._5B5&c=~-~-~&f=5l&p=1&a=5&y=334&s=0&r=rs&version=2.0.1 Shouldn't `4\5 + 400.` come after `5\5`?

Intuitively yes, but the last interval in the scale is special. This modifier should not touch the equave/period.

frostburn:
**Sevish** — Today at 7:00 PM

> it would be nice to have that option that keeping the offset is more important than the equave
> for example if I wanted to offset the scale by 2/1 to basically double up what I had before, so I can apply a change to the 2nd octave
> or maybe someone will want to have the equave changed so they can do some other process before reducing

frostburn:
Changed the "Keep" option to modify the equave when overflow happens.

Issue #375, opened 2022-12-08
Clarify MIDI mappings with pictures


frostburn:
Craft SVG diagrams that show how the standard piano layout maps to scale degrees.

PR #390, opened 2022-12-19
Visualize MIDI mapping and active MIDI keys

frostburn:
Add a missing index to the computed white indices

ref #375


frostburn:
Self-reviewing: Here goes nothing!

Issue #397, opened 2022-12-31
Add support for tempering in huge subgroups


frostburn:
**Userminusone** — 12/24/2022 5:39 AM

> So it turns out ScaleWorkshop tempering via commas just... fails with high prime limits?
>
> Is this a bug or am I not supposed to go up to the 31 limit with tempering?

This is related to: https://github.com/xenharmonic-devs/temperaments/issues/30

frostburn:
I think the best I can do for now is making commas disappear via generic gradient descent and adding vals in Tenney-Euclid optimal linear combinations. I should be able to keep equaves pure, but not support arbitrary constraints.

frostburn:
Note that this doesn't apply to scale generation. Those remain off-limits for now.

You need the wedgie to work out the period and the generator. The naive way of calculating the wedgie grows like `2**n` in complexity. There might be a `n**2` solution, but I haven't figured it out yet.

The numerical methods for working out periods and generators from mapping vectors are somewhat flimsy. Producing the wrong result or randomly failing is not great for UX.

frostburn:
I guess I'll implement everything in this repository for now. Can't see a reasonable way around adding `mathjs` dependency and don't want to burden `temperaments` with it.

PR #398, opened 2023-01-02
Implement lightweight tempering algorithms for large subgroups

frostburn:
ref #397


frostburn:
Went ahead and included `mathjs` higher upstream as it could simplify some matrix calculations there anyway.

frostburn:
I need to look into this and describe how to test.

frostburn:
To test load a JI preset like *Raga Bageshri* and apply the *Temper* modifier with prime limit set to something high like 41. In `main` that eventually crashes something, but in this branch it should work.

frostburn:
Self-reviewing: For the love of Spoob, why are we like this?!

PR #408, opened 2023-01-29
Stop Firefox from bringing up Quick Find when '/' is pressed

### LINKED ISSUE #407: Stop Firefox from opening Quick Find when '/' key is pressed
Firefox has a feature called "Quick Find" which makes the slash key bring up a search bar unless text is being entered. This interferes with the using the keyboard for notes in Scale Workshop.

A fix has been provided.

forrcaho:
Fixes #407

### DIFF
--- a/src/keyboard.ts
+++ b/src/keyboard.ts
@@ -223,6 +223,7 @@ export class Keyboard {
     }
 
     if (COORDS_BY_CODE.has(event.code)) {
+      event.preventDefault(); // Stops Firefox from intercepting '/' for Quick Find
       this.activeKeys.add(event.code);
       if (event.shiftKey) {
         this.pendingKeys.add(event.code);


frostburn:
I find this solution too heavy-handed.  It's not the job of a key-press wrapper to interfere with default browser behavior. Maybe `App.windowKeydown` is a better place for this.

forrcaho:
This fix is based on [a StackOverflow answer](https://stackoverflow.com/a/8563788). Notice that it has to check that the focus is not on an input element before reclaiming the key event from Firefox.

By inserting the line where I did, the focus state is known: any keydown event executing that code is intended to play a note. 

I'm not familiar with `App.windowKeydown`, but I worry that putting this code elsewhere would mean that the focus check would have to be duplicated.

frostburn:
I'm pretty sure this fix can be placed near the focus check without having to worry about code duplication. I can take a crack at it later. The fix itself is welcome. :pray:

frostburn:
Rejected in favor of #409

PR #409, opened 2023-02-01
Disable browser specific features like quick find on Firefox

frostburn:
ref #407


frostburn:
Intended to replace #408

SeanArchibald (approved):

PR #421, opened 2023-04-25
Document changes from 2.0.1 to 2.1.0 in the README

frostburn:


frostburn:
There are more than a couple of features here. How about bumping the version all the way to `2.1.0` instead?

SeanArchibald:
I think 2.1.0 would be fine

SeanArchibald (approved):

Issue #416, opened 2023-04-16
Passing float for MIDI note number sometimes causes black screen


rlh74:
Accidentally noticed this when I thought I was setting base frequency input field but instead was focused on MIDI note for base frequency input field. Any decimal values other than 0 will cause the screen to go black, clicking somewhere in the browser window will bring the page back with the scale output table missing. When running in local development environment, there is no UI bug but the MIDI note number displays undefined for midi note name and a float for the octave number. 

Updating emit args for `baseMidiNote` (line 103 in `ScaleBuilder.vue`) to `Math.trunc(newValue)` provides a quick fix that handles decimal values the same way other invalid characters are handled. Unit tests pass and e2e tests appear unaffected by the change.

PR #422, opened 2023-04-27
Force base MIDI note to be an integer

frostburn:
ref #416


frostburn:

Issue #423, opened 2023-04-28
Invalid lines in the URL get converted into garbage


frostburn:
Invalid scale lines like plain words get passed to the URL but are decoded as numbers instead.

PR #424, opened 2023-04-30
Escape invalid lines in the URL

frostburn:
ref #423


frostburn:
Self-reviewing: I guess we have a horrible way of adding comments in the scale now.

Issue #401, opened 2023-01-07
Rename standard reference pitch to A4 instead of A5


frostburn:
**fredg999** — _Today at 4:34 AM_

> In the "Build Scale" view, under "Tuning", the label next to the "MIDI note for base frequency" input box says "A5" for a value of 69 (default), and moves accordingly when I go up and down. However, correct me if I'm wrong, but shouldn't that be A4?
> https://en.wikipedia.org/wiki/A440_(pitch_standard)
> I remember being confused about octave notation when I was younger because French musicians used octave 5 when English musicians would use 4, but it seems like the English version is the most common one nowadays.
> https://en.wikipedia.org/wiki/Scientific_pitch_notation

KoenTanghe:
Just discovered scale workshop today (what a nice tool!) and this also confused me.

So I would personally also prefer to see the standard reference pitch as "A4" (note 69, standard 440), as the Scientific pitch notation (SPN) / American standard pitch notation (ASPN) / International Pitch Notation (IPN) defines it. https://en.wikipedia.org/wiki/Scientific_pitch_notation

frostburn:
> Just discovered scale workshop today (what a nice tool!) and this also confused me.
> 
> So I would personally also prefer to see the standard reference pitch as "A4" (note 69, standard 440), as the Scientific pitch notation (SPN) / American standard pitch notation (ASPN) / International Pitch Notation (IPN) defines it. https://en.wikipedia.org/wiki/Scientific_pitch_notation

Glad you like it! We're currently short on volunteers. Someone just has to look at #426 and tell it's OK to get the issue is fixed.

000masa000:
I will look at this as soon as I can.... seems like it is good to implement

PR #426, opened 2023-05-02
Make A4 the default reference pitch

frostburn:
Use the English standard for 69 = A4 by default.
Add configuration to recover the old French behavior 69 = A5.

ref #401


frostburn:

Issue #394, opened 2022-12-28
Touching the same key twice leaves yellow marks on virtual keyboard


frostburn:
**To reproduce:**

1. Touch a key on the virtual keyboard
2. Drag finger away from the key
3. Touch the same key with another finger
4. Release both fingers

**Result:** The note keeps ringing and a yellow mark is left on the keyboard

**Expected result:** The note stops ringing with no residue.

PR #428, opened 2023-05-02
Prevent multiple notes on a single virtual key

frostburn:
ref #394


frostburn:

Issue #362, opened 2022-11-28
Long Chord Wheel chords do not fit inside canvas


frostburn:
**Dead Shaman** — Yesterday at 11:18 PM

> ...
> Numbers sometimes don't fit inside the box.
> ...
![image](https://user-images.githubusercontent.com/1253499/204216212-38541f86-953b-483a-8ae0-5ce7ad75686e.png)

PR #429, opened 2023-05-02
Adjust chord wheel text size dynamically to fit

frostburn:
ref #362


frostburn:

Issue #378, opened 2022-12-10
Rotation ignores precision preferences


frostburn:
[Weltfried](https://www.reddit.com/user/Weltfried/) [on reddit](https://www.reddit.com/r/musictheory/comments/zexkrm/comment/izixtwl/?utm_source=reddit&utm_medium=web2x&context=3)

Try rotating:
```
5/4 - 0.31
4/3 + 7.797
3/2 + 2.095
2/1 + 6.595
```

PR #430, opened 2023-05-02
Attach default interval options to parsed scale lines

frostburn:
ref #378


frostburn:

Issue #420, opened 2023-04-21
Sorting causes composite intervals to be respelled


frostburn:
Composite intervals get respelled when sorting e.g. `13/8 - 3/2` becomes `13/12` , making them harder to recognize and losing information about the intention.

frostburn:
Core functionality added: https://github.com/xenharmonic-devs/scale-workshop-core/commit/4abed1febc5177a955a528b500a8618ee0ea4f52

PR #431, opened 2023-05-03
Use name preserving reverse parsing when modifying scales

frostburn:
ref #420


frostburn:

Issue #412, opened 2023-03-25
Interval matrix shows 1/1 when mixing cents and ratios


frostburn:
**fredg999** — Today at 5:39 AM

> I think I found a bug in SW's interval matrix (Analysis tab). Here's an example of scale that triggers it: https://sevish.com/scaleworkshop/analysis?l=5k._b4._dw._jg._p0._uk._2F1&version=2.0.1
> The "default" display shows 1/1 for a bunch of intervals (roughly the top-left half, diagonally), whereas the "cents" and "decimal ratio"  displays show the correct values in each cell.
> The bug seems to be associated with the mixed use of cents and ratios (2/1 for the octave, but cents for the rest). Replacing 2/1 with 1200. is a possible workaround, but I know that many Scala files (as found in the Scala scale archive, for example) are formatted like this, with many cents values and then a 2/1 for the octave, so I think this bug might be encountered by other people under "normal" circumstances.

frostburn:
I think this was an issue in the core where intervals within the scale got 1/1 added by accident: https://github.com/xenharmonic-devs/scale-workshop-core/commit/a24624d5a44adb421705089b038bf1cdc2eb771b

frostburn:
Still needs a minor tweak in the core to handle the root mode better.

PR #433, opened 2023-05-03
Always show accurate information in the interval matrix 

frostburn:
ref #412


frostburn:

Issue #356, opened 2022-11-27
Reverb


budjarnlambeth:
Would it be possible to add a reverb effect to the synth, the same way you could in the previous version of Scale Workshop?

frostburn:
It certainly would be. Adding this as a feature request.
By the way, the previous version is still available if you miss it: https://sevish.com/scaleworkshop1/

PR #439, opened 2023-05-19
Implement ping pong delay for basic reverb

frostburn:
ref #356


frostburn:
Self-reviewing: Sounds nice.

Issue #403, opened 2023-01-21
Expand SI units to include everything from quecto to quetta.


frostburn:
https://en.wikipedia.org/wiki/Metric_prefix

PR #440, opened 2023-05-19
Expand SI units to include everything from quecto to quetta

frostburn:
ref #403


frostburn:
Self-reviewing: Happy physicsing! :atom_symbol:

Issue #364, opened 2022-11-28
Some modals inputs lack monzo support


frostburn:
**Dead Shaman** — Today at 12:03 PM

> _New scale - Span lattice_
> and
> _New scale - Combination product set_
> don't allow to write Generators using monzos. 
> Not a very severe issue, you can still get the desired results writing everything as cents. But, an issue nevertheless.

frostburn:
Somewhat related to #177

frostburn:
Core functionality implemented: https://github.com/xenharmonic-devs/scale-workshop-core/commit/c89c8e7efbb16bf41dd3fbde979e6b8db1bf011b

PR #441, opened 2023-05-21
Add support for monzo syntax inside chords and comma lists

frostburn:
refs #364 #177


frostburn:
The `scale-workshop-core` dependency needs to be converted to a proper release tag before this can be merged.

frostburn:
Self-reviewing: Seems to do what it's supposed to.

Issue #338, opened 2022-11-11
Stretch into a specific interval


frostburn:
**0.5° 🌜** — Today at 2:42 PM

> [...] what could be added to Stretch/compress to make it comfortable to use [...]? Maybe adding fields what interval to stretch into what? Like “make 8\13 into 3/2” will set stretch ratio to ln(3/2) / ln(8\13) ≈ 0.951

**0.5° 🌜** — Today at 2:44 PM

> it could be “take an interval from M-th to N-th degree of the current scale and make it into this interval” but that seems cumbersome, given there is an interval matrix to see which interval is that. So I think the first idea is better

PR #443, opened 2023-05-28
Add support for stretching to a specific interval

frostburn:
ref #338


frostburn:
Self-reviewing: Seems to work as intended.

Issue #387, opened 2022-12-17
Feature request: exporting scales via MIDI SysEx for the Audiothingies MicroMonsta 2


tobia:
The [Audiothingies MicroMonsta 2](https://www.audiothingies.com/product/micromonsta2/) is a small and relatively cheap digital polyphonic hardware synth that, among other things, supports MPE and custom tuning.

MPE allows for each note to be pitch shifted and modulated in real-time via MPE-compatible controllers, such as the ROLI Seaboard or Roger Linn's LinnStrument. Custom tuning allows mapping the 128 MIDI notes to any custom scale.

I personally bought this synth for these two features, but the device does not have a user interface to input scales, they must be sent over MIDI and are lost when powering it down. 

Here is what [the manual](https://www.audiothingies.com/dl/micromonsta_2/docs/MicroMonsta%202%20Manual%20EN%202.3.pdf) has to say (Microtuning, page 34):

> Micromonsta 2 can be retuned globally by sending a Bulk Tuning Dump via MIDI.
> A bulk tuning dump has the following format:
> `F0 7E <device ID> 08 01 tt <tuning name> [xx yy zz] ... chksum F7`
> where:
> - `<device ID>` is 0x00 or 0x7F (all devices)
> - `tt` is tuning program number (not used, set to 0x00)
> - `<tuning name>` is 16 ASCII characters (not used, set it to “NotUsedButNeeded” for example)
> - `[xx yy zz]` is the frequency data for one note (repeated 128 times)
> - `chksum` is the calculated checksum (not used)
>
> Bulk Tuning Dumps are global, all the patches will be affected.
> They are not stored in Micromonsta 2 memory, this means that each time you power it off, it will lose your custom tuning, and you will need to send it again for the working session.

Would it be possible to use Scale Workshop's MIDI capabilities to send a scale directly to this synth by crafting the appropriate SysEx message?

tobia:
Slightly OT: [here's my favorite demo](https://www.youtube.com/watch?v=d3NqpzAARvU) (not mine) of this small synth. It's not using MPE nor custom scales, but I find the sound quality quite impressive.

frostburn:
Scale Workshop uses [WebMidi](https://webmidijs.org/) under the hood and it should be capable of sending SysEx messages. (I'm pretty sure we could still use the browser MIDI API directly if WebMidi fails.)

[...]

Issue #384, opened 2022-12-15
Feature request: add export to Minilogue XD format


znmeb:
I have a Minilogue XD and would like to use Scale Workshop to create user scale (`.mnlgxdtuns`) files for it. I tried using a `.mnlgtuns` file from Scale Workshop and renaming it, but the Minilogue XD Sound Librarian throws an error "File is not a minilogue xd data file.".

I haven't been able to find online documentation on either file format. What I do have, though, is the MIDI specification for the Minilogue XD (from <https://www.korg.com/us/support/download/manual/0/811/4440/> (attached).

Also, I have tried loading the same scale (Eikosany 1-3-5-7-9-11) using `.scl` and `.kbm` files. That doesn't work either. The Sound Librarian appears to be ignoring the `.kbm` file and remapping the keyboard starting at MIDI note number 0, which is two octaves lower than the normal mapping. I've asked Korg support for help with this.

[minilogue_xd__MIDIImp.txt](https://github.com/xenharmonic-devs/scale-workshop/files/10233230/minilogue_xd__MIDIImp.txt)

vsicurella:
The differences between the formats are extremely trivial - a bit annoying they chose to use a different file type for every synth model! 
I can concur with your observations on the KBM file too, it seems like they literally do not pay any attention to it at all despite providing an import for it.

The "scale" format can work around this by just centering the tuning values around the root that's chosen in SW, however it's not straightforward for the "octave" format. The octave format is bounded by the lowest possible note being defined as C, which makes it so if you want the tuning to be consistent you have to transpose it be rooted on standard C tuning. I don't have a better alternative for that now.

Anyway, I propose we make the Korg exporter a modal where you can choose the specific 'logue synth model.

PR #447, opened 2023-07-08
MTS "Bulk Tuning Dump" exporter & Korg exporter refactoring

vsicurella:
This adds a .syx exporter that adheres to the MIDI Tuning specification of the "Bulk Tuning Dump" SysEx message, and also refactors the Korg exporter to be able to export in different formats associated with different supported synth models. 

The Korg exporter now relies on some of the same code as the MTS Exporter since it's really just a wrapper around the same data. 

Note: The previous Korg exporter implementation had some math errors affecting the least-signficant byte of the tuning data for each note. I'm not sure the extent of the error (it's a tricky calculation) but I would expect noticeable differences and would encourage users to re-export and re-import their scales.


src/components/modals/export/MtsSysexExport.vue
```diff
@@ -0,0 +1,116 @@
...
+    </template>
+  </Modal>
+</template>
+
+<style>
+input.half {
+  flex-grow: 0.25 !important;
+}
+
+span.info-question {
```
frostburn, line 99: This seems useful. Wouldn't mind having it in the main style file so that we can implement more tooltips.

src/components/modals/export/MtsSysexExport.vue
```diff
@@ -0,0 +1,116 @@
...
+    <template #body>
+      <div class="control-group">
+        <div class="control">
+          <label for="name">Name (16-character limit)</label>
+          <input
+            class="half"
+            type="text"
+            id="name"
+            v-model="name"
+            maxlength="16"
```
frostburn, line 67: Are you sure this is enough validation? I'm unable to break it myself, but you could enforce `clampName` everytime the ref is updated.

frostburn:
The commit messages should reference their respective issues #384 and #387 .

frostburn:
The sister PR https://github.com/xenharmonic-devs/xen-dev-utils/pull/7 needs to be resolved and published before this PR can be merged.

src/components/modals/export/MtsSysexExport.vue
```diff
@@ -0,0 +1,116 @@
...
+        </div>
+        <div class="control">
+          <label for="preset-index">
+            Preset Index&nbsp;
+            <span
+              @click="$event.preventDefault()"
+              class="info-question"
+              title="Refer to your synth's manual for a valid range"
+            >
+              ?
```
frostburn, line 78: This question mark could be moved to a CSS `content` property.

src/components/modals/export/MtsSysexExport.vue
vsicurella: That must have been an oversight after I realized how to do it for presetIndex. I'm used to using computed properties in Vue2 for this kinda stuff, so I was toying around with it before I figured it out.

src/components/modals/export/MtsSysexExport.vue
```diff
@@ -0,0 +1,91 @@
...
+    typeof input === "string" ? parseInt(input.replace(/\D/g, "")) : input;
+  if (Number.isNaN(number)) return 0;
+  return clamp(0, 127, number);
+}
+
+// Rarely implemented parameters
+// const deviceId = ref(0);
+// const bank = ref(0); (only for message 0x0804)
+
+const name = ref(clampName(props.scaleName));
```
frostburn, line 33: This update should hapen every time the modal is opened. Now this line of code basically reads like a statement of intent that masquerades as a computed property.

vsicurella: I was able to add a watcher on `props` to solve this, but I'm not sure if this is the best way to implement input formatting. Do you think there's a better way? 
I tried an approach with a writable computed property (something I normally do in BootstrapVue) but couldn't get it to work, plus the Vue3 documentation says that should only be used in rare cases anyway.

frostburn: Yeah, it's a weird case because you want changes in the scale name to affect the clamped name but not the other way around. I'll check if there's a straightforward solution.

frostburn: This does what I'm after: Updating the name on the main interface changes what's in the modal, but modifying the modal text input doesn't reflect back on the main interface.
```Javascript
const name = ref("");

watch(
  () => props.scaleName,
  (newValue) => {
    name.value = clampName(newValue);
  },
  { immediate: true }
);
```

vsicurella: Thanks so much! That's close to what I came up with too, good to have some validation on that.

vsicurella:
The one thing I noticed that I'd rather pass on fully solving for this particular PR is improving the circle consistency of the info-question CSS definition. I noticed on a different computer it doesn't render a perfect circle, I'm guessing because of OS font differences.

frostburn:
> The one thing I noticed that I'd rather pass on fully solving for this particular PR is improving the circle consistency of the info-question CSS definition. I noticed on a different computer it doesn't render a perfect circle, I'm guessing because of OS font differences.

That's perfectly OK. Simply create a new issue about circle consistency and reference this PR so that we can fix it later.

src/components/modals/export/KorgExport.vue
```diff
@@ -0,0 +1,109 @@
...
+    filename: sanitizeFilename(props.scaleName),
+    baseMidiNote: props.baseMidiNote,
+  };
+
+  const exporter = new KorgExporter(
+    params,
+    modelName.value,
+    useOctaveFormat.value
+  );
+  exporter.saveFile();
```
frostburn, line 54: await exporter.saveFile();

src/components/modals/export/KorgExport.vue
```diff
@@ -0,0 +1,109 @@
...
+  // Can check for other errors here...
+  return String();
+});
+
+const fileTypePreview = computed(() => {
+  const format = getKorgModelInfo(modelName.value);
+  return useOctaveFormat.value ? format.octave : format.scale;
+});
+
+function doExport() {
```
frostburn, line 41: `async function doExport() {`

src/components/modals/export/KorgExport.vue
```diff
@@ -0,0 +1,109 @@
...
+          <p class="alert-message-danger">
+            {{ dialogErrorMessage }}
+          </p>
+        </div>
+      </div>
+    </template>
+    <template #footer>
+      <div class="btn-group">
+        <button
+          @click="$emit('confirm')"
```
frostburn, line 100: `@click="doExport"`

src/assets/base.css
```diff
@@ -274,3 +282,39 @@ p.section {
...
+  background: var(--color-alert-danger-background);
+  border-width: 1px;
+  border-radius: 1%;
+  border-style: solid;
+  border-color: var(--color-alert-danger-border);
+  padding: 8px;
+}
+p.alert-message-danger {
+  color: var(--color-alert-danger);
+}
```
frostburn, line 320: Missing newline at the end of the file.

frostburn:
The 'OK' button must be fixed. As of now clicking it does nothing.

Other than that, great work!

I won't test the actual files generated as I don't have the hardware. We have two production deployments so Korg owners can do the testing for us.

src/components/modals/export/KorgExport.vue
vsicurella: embarrassing...thank you lol

src/components/modals/export/KorgExport.vue
```diff
@@ -0,0 +1,109 @@
...
```
frostburn, line 100: Still not calling `doExport`.

frostburn:
Seems gnome :elf: free now. Merging.

Issue #446, opened 2023-06-27
TUN file import doesn't implement specified base frequency


mousemambo:
Importing a .tun file reads and adds to the page the "MIDI note for base frequency" from the .tun file, but NOT the "Base frequency" leaving it as whatever it happened to be set to before the import (440 by default on a fresh page, or whatever was previously loaded). As a result, the calculated table is usually very wrong after import.

  To reproduce, start with a fresh page ( https://sevish.com/scaleworkshop/?version=2.1.0 );
  Notice the base freq. is 440 and the MIDI note is 69;
  Load a .tun file whose tuning center is not 440/69 (e.g. attached file "HarrisonM-RagaJust-C.tun" which was created using Scale Workshop);
  Notice that the MIDI note has been updated to what is in the .tun file (60 in this case), but the base freq. is still 440 (should be 261.6256 as specified by the .tun file;

If you then hit "Auto" next to the Base Freq. field, the correct value is determined and shown. But I don't see any reason the import process shouldn't use the Base Freq. explicitly given in the .tun file. Adding the extra step seems obscure, undocumented, and unnecessary, so I suspect this is a bug. Thank you for continuing to develop this great tool!

[HarrisonM-RagaJust-C.zip](https://github.com/xenharmonic-devs/scale-workshop/files/11877909/HarrisonM-RagaJust-C.zip)

mousemambo:
Thank you for fixing this! I look forward to the next release.

PR #449, opened 2023-07-16
Import base frequency from Anamark v2 tunings

frostburn:
ref #446


frostburn:
Self-reviewing: Seems to work.

Issue #38, opened 2022-05-14
Improved Lattice Visualization


frostburn:
Visualize the lattice of just intonation so that similar intervals have similar angles and lengths in the visualization.
Use comma vectors to make the visualization periodic.

frostburn:
I think an `<svg>` element is a good fit for this. I wouldn't go as far as using Vue for the child elements because lattices can have hundreds of elements that we want to interact with. I suspect that isometric orthogonal projection is enough to make a useful tool for displaying JI relationships and even for adding new notes to the scale. While a bit clumsy, an all-positive Clifford algebra might suffice and those are available through `temperaments.getAlgebra(n, 'float64')`. Sharing a cache with tempering saves memory too.

frostburn:
An interactive lattice is more work. We can start with a plain visualization in the *Analysis* tab with draggable prime vectors and no need to worry about orthonormality and algebras.
To make dragging less compute-intensive we can drop periodicity while dragging the mapping vectors.

frostburn:
So we're finally moving ahead with a lattice in #453, features still missing:

- [ ] More prime vectors #455
- [ ] Support for equaves that are a power of a prime (only requires slicing of the monzos)
- [ ] Support for any equave (requires construction of a new orthonormal subspace to replace the affected primes)
- [ ] Support for user-defined prime vectors
- [ ] Repetition of the scale in the direction of user-defined commas
- [ ] Refactor the code so that node&edge creation is a separate unit-tested utility
- [ ] *(optional)* "perspective" (make some primes cause the whole visual to approach a vanishing point)
- [ ] *(experimental)* see if fractions of the octave can be automatically detempered into a 5-limit lattice for visualization or if there's some other natural way to make edos 2d.

frostburn:
**richie** — _Today at 6:19 PM_

> I spent a couple hours using this device last night for a project I'm working on! Thanks so much!
> These ideas are probably all well considered so I don't mean for'm to be boring/bother. Two things that come to mind:
> I'd love to have an option/way to change the size of the text, even potentially hide it altogether if desired.
> on several of the lattices I was making, text was quickly overlapping/cluttered feeling; tried to use the "zoom" function in my brower to no avail.
> When removing altogether I could see it as an inviting way for people to demonstrate their scales in videos too, given the interactive "light" up vertices. (nice touch!)

frostburn:
**sintel 🎏** — _Today at 7:04 PM_

[...]

PR #453, opened 2023-10-09
23-limit SVG lattice

diegovdc:


src/components/ScaleLattice.vue
```diff
@@ -1,19 +1,195 @@
 otonalFundamental
 <script setup lang="ts">
-import type { Scale } from "scale-workshop-core";
-import { ref } from "vue";
-
-defineProps<{
+import { access } from "fs";
```
frostburn, line 3: Huh?

src/components/ScaleLattice.vue
```diff
@@ -1,19 +1,195 @@
...
+    for (let j = i + 1; j < arr.length; j++) {
+      combinations.push([arr[i], arr[j]]);
+    }
+  }
+
+  return combinations;
+}
+
+// TODO move somewhere else
+function euclideanDistance(point1: number[], point2: number[]): number {
```
frostburn, line 78: You might want to look into `Math.hypot`. IIRC it supports multiple arguments.

src/components/ScaleLattice.vue
```diff
@@ -1,19 +1,195 @@
...
+
+// === Computed state ===
+const equave = 2; // TODO unassume equave is 2
+const edges = computed(() => {
+  const combinations = getPairCombinations(props.scale.intervals);
+  return combinations
+    .map(([a, b]) => {
+      return {
+        diff: euclideanDistance(
+          a.monzo.vector.valueOf().slice(1),
```
frostburn, line 99: `a.monzo.toIntegerMonzo().slice(1)` instead.

src/components/ScaleLattice.vue
```diff
@@ -1,19 +1,195 @@
...
+        :y2="nodesByLabel[edge.pair[1]].y"
+        stroke="white"
+        stroke-width="1"
+      />
+      <g v-for="node of nodes" :key="node.label">
+        <circle :cx="node.x" :cy="node.y" r="4" fill="white" />
+        <text
+          :x="node.x + 2"
+          :y="node.y - 4"
+          font-family="Arial"
```
frostburn, line 186: Stuff like fonts and fill colors should be in a `<style>` section and the elements should only get a class to reference the styles.

src/components/ScaleLattice.vue
diegovdc: VSCode autoimport (didn't clean this up), lol

src/components/ScaleLattice.vue
```diff
@@ -1,19 +1,282 @@
...
+    (
+      primeIndex: number | null,
+      component: number,
+      i: number
+    ): number | null => {
+      if (component === 1 && !primeIndex) {
+        return i;
+      }
+      if (component !== 0) {
+        console.error("Equave is not a prime number");
```
frostburn, line 100: An error in the console is too obscure if this is intended to be seen by the user.

src/components/ScaleLattice.vue
```diff
@@ -1,19 +1,282 @@
...
+        .reduce(
+          (isZero: boolean, component: number) => isZero && component === 0,
+          true
+        ),
+    true
+  )
+);
+
+function maybeGetPrimeEquave(equave: Interval): number | null {
+  const intmonzo: number[] = equave.monzo.toIntegerMonzo();
```
frostburn, line 89: Bad variable name. Prefer `integerMonzo`.

src/components/ScaleLattice.vue
```diff
@@ -1,19 +1,282 @@
...
+              y: y + vector.y * componentVal,
+            };
+          }
+        },
+        { x: 0, y: 0, label, index, fill: "white" }
+      );
+      return node;
+    });
+  } catch {
+    return new Error("Cannot make lattice of non JI scale.");
```
frostburn, line 141: You `throw` errors, not `return` them.

src/components/ScaleLattice.vue
```diff
@@ -1,19 +1,282 @@
...
+
+const edges = computed(() => {
+  if (!Array.isArray(nodes.value)) {
+    return [];
+  }
+  const combinations = getPairCombinations(props.scale.intervals);
+  return combinations
+    .map(([a, b]) => {
+      return {
+        diff: euclideanDistance(
```
frostburn, line 159: Poor field name. Mathematically not a difference but a distance instead.

frostburn:
Crashes with an unrecoverable error when trying to load the "Slendro" preset in the lattice tab.

src/components/ScaleLattice.vue
```diff
@@ -1,19 +1,282 @@
...
 </script>
 
 <template>
-  <canvas ref="canvas" :width="width" :height="height"> </canvas>
+  <div>
+    <div v-if="error">
+      <p>{{ error }}</p>
+    </div>
+    <div v-else>
+      <svg
```
frostburn, line 210: Dynamic graphics like this should have an explicit `preserveAspectRatio`.

src/utils.ts
```diff
@@ -238,3 +238,26 @@ export function computedAndError<T>(
...
+    for (let j = i + 1; j < arr.length; j++) {
+      combinations.push([arr[i], arr[j]]);
+    }
+  }
+
+  return combinations;
+}
+
+export function euclideanDistance(point1: number[], point2: number[]): number {
+  if (point1.length !== point2.length) {
```
frostburn, line 255: In monzo context you can assume that the missing coordinates are all zero.

frostburn:
Should target `main` instead of `lattice-stub`. There's some churn (such as canvas being replaced with svg) compared to my preliminary work. Please erase it and start fresh so the commit history looks clean.

frostburn:
Make sure the lattice looks good on smartphone screens too. I would add a button to rotate everything (except the labels) 90 degrees to better make use of vertical space when available.

src/components/ScaleLattice.vue
frostburn: (Maybe this was part of some other machinery. Just came home, didn't look too deep. In any case it's not good practice to repurpose the `Error` class for other uses. Make a similar custom class if you must.)

src/components/ScaleLattice.vue
diegovdc: It was not intended to be seen by the user, but any suggestion on how to improve this is welcome.

diegovdc:
> Make sure the lattice looks good on smartphone screens too. I would add a button to rotate everything (except the labels) 90 degrees to better make use of vertical space when available.

I think it's better for users to be able to rotate their mobile, rotating the coords can be very confusing in my experience.

frostburn:
The visuals have poor legibility when the labels cross with the lattice edges:
image
The labels should stand out or avoid the edges. I suggest adding a stroke that matches the background color.

frostburn:
The visualization blows up when used with an empty scale:
image

src/components/ScaleLattice.vue
```diff
@@ -0,0 +1,340 @@
+otonalFundamental
```
frostburn, line 1: What's this?

diegovdc: No idea, it comes from your branch.

src/views/LatticeView.vue
```diff
@@ -0,0 +1,52 @@
+<script setup lang="ts">
+import { computed, ref } from "vue";
+import type { Scale } from "scale-workshop-core";
+import { mmod } from "xen-dev-utils";
+import ScaleLattice from "@/components/ScaleLattice.vue";
+
+// See ji-lattice-2 branch for ideas and how to extract coordinates from the scale.
```
frostburn, line 7: These comments were inteded as guidance during development. They should not end up in the main branch. If you need help with the bootstrapping I can take over the branch once everything else has been dealt with.

diegovdc:
> The visualization blows up when used with an empty scale: image

Why does an empty scale contains a 1/1 interval?

src/views/LatticeView.vue
```diff
@@ -0,0 +1,52 @@
...
+  keyColors: string[];
+  heldNotes: Map<number, number>;
+}>();
+
+// Please make responsive or something
+const canvasWidth = ref(300);
+const canvasHeight = ref(200);
+
+// Use these to light up degrees of the scale in the lattice
+const heldScaleDegrees = computed(() => {
```
frostburn, line 20: This computed property is broken but the intended functionality of lighting up the lattice nodes is desirable. I can fix the property tomorrow...

src/components/ScaleLattice.vue
```diff
@@ -0,0 +1,340 @@
...
+    }
+  }
+};
+
+onMounted(() => {
+  adjustLatticeSize();
+  window.addEventListener("resize", adjustLatticeSize);
+});
+
+onBeforeUnmount(() => {
```
frostburn, line 103: Any particular reason why this isn't the more idiomatic `onUnmounted`?

src/components/ScaleLattice.vue
```diff
@@ -0,0 +1,340 @@
...
+  type ComputedRef,
+  ref,
+  onBeforeUnmount,
+} from "vue";
+import { PRIMES } from "xen-dev-utils";
+import { monzoEuclideanDistance, getPairCombinations } from "@/utils";
+const props = defineProps<{
+  scale: Scale;
+  width: number;
+  height: number;
```
frostburn, line 16: Unused props.

src/views/LatticeView.vue
```diff
@@ -0,0 +1,52 @@
...
+
+<template>
+  <main>
+    <h2>Lattice visualization</h2>
+    <ScaleLattice
+      :scale="scale"
+      :width="canvasWidth"
+      :height="canvasHeight"
+      :heldScaleDegrees="heldScaleDegrees"
+      :keyColors="keyColors"
```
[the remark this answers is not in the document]
diegovdc: yep

src/components/ScaleLattice.vue
```diff
@@ -0,0 +1,340 @@
...
+  },
+  "23": {
+    x: 20,
+    y: -6,
+  },
+};
+
+const container = ref<HTMLDivElement | null>(null);
+
+const adjustLatticeSize = () => {
```
frostburn, line 78: I'm inclined to believe that these acrobatics can be avoided by setting the `<svg>`'s `width` and `height` attributes to `100%` and carefully setting the `viewBox` and `preserveAspectRatio` attributes with respect to the coords or vice versa.
Not a blocker though. Can be deferred to a separate issue to be solved later.

diegovdc: Yeah, I am not an svg expert so if you know how to do it I would appreciate some more concrete suggestions.

src/components/ScaleLattice.vue
```diff
@@ -0,0 +1,340 @@
...
+const viewBox = computed(() => {
+  const defaultViewBox = { minX: 0, minY: 0, maxX: 0, maxY: 0 };
+  if (svgElement.value) {
+    return [
+      ...svgElement.value.querySelectorAll("text"),
+      ...svgElement.value.querySelectorAll("circle"),
+    ].reduce(({ minX, maxX, minY, maxY }, el) => {
+      const { x, y, width, height } = el.getBBox();
+      return {
+        minX: minX < x - width / 2 ? minX : x - width / 2,
```
frostburn, line 229: `Math.min` is more idiomatic JS in my opinion.

src/components/ScaleLattice.vue
```diff
@@ -0,0 +1,340 @@
...
+      .filter(({ distance }) => distance === 1);
+  } catch {
+    return [];
+  }
+});
+
+const svgElement = ref<HTMLElement | null>(null);
+
+const viewBox = computed(() => {
+  const defaultViewBox = { minX: 0, minY: 0, maxX: 0, maxY: 0 };
```
frostburn, line 221: It's better to start with minima set to `Infinity` and maxima set to `-Infinity` before reducing.

src/components/ScaleLattice.vue
```diff
@@ -0,0 +1,340 @@
...
+          class="edge"
+          :key="edge.pair.join()"
+          :x1="nodesByLabel[edge.pair[0]].x"
+          :x2="nodesByLabel[edge.pair[1]].x"
+          :y1="nodesByLabel[edge.pair[0]].y"
+          :y2="nodesByLabel[edge.pair[1]].y"
+        />
+        <circle
+          v-for="node of nodes"
+          :key="node.label"
```
frostburn, line 291: Key not unique when the scale contains duplicate entries.

src/components/ScaleLattice.vue
```diff
@@ -0,0 +1,340 @@
...
+          :cy="node.y"
+          r="2.5"
+          :class="`node   ${
+            props.heldScaleDegrees.has(node.index) ? 'heldDegree' : ''
+          }`"
+        />
+        <!-- Separating the iteration of the labels so that circles don't overlap any text -->
+        <text
+          v-for="node of nodes"
+          :key="node.label + 'text'"
```
frostburn, line 302: See issue about duplicity above. Keys only need to be unique within the `v-for` so `+ 'text'` is superfluous here.

src/components/ScaleLattice.vue
```diff
@@ -0,0 +1,340 @@
...
+  label: string;
+  index: number;
+  fill: string;
+};
+
+type Coords = {
+  [key: string]: Coord;
+};
+
+const coords: Coords = {
```
frostburn, line 33: I'd prefer this to be an array corresponding to monzo indices with comments indicating which prime we're dealing with. There's no need to map monzo indices to primes to strings to object keys.

src/components/ScaleLattice.vue
```diff
@@ -0,0 +1,340 @@
...
+  },
+  "17": {
+    x: -5,
+    y: -32,
+  },
+  "19": {
+    x: 7,
+    y: -25,
+  },
+  "23": {
```
frostburn, line 70: Scale Workshop goes up to 97. Please make an issue about supporting the full range if you don't want to do all primes in this PR.

src/components/ScaleLattice.vue
frostburn: Oops, my bad! :sweat_smile: Can be removed.

frostburn:
> > The visualization blows up when used with an empty scale: image
> 
> Why does an empty scale contain a 1/1 interval?

The scale has to start from somewhere even if it doesn't end on anything. The empty scale is an anomaly, but still valid.

src/components/ScaleLattice.vue
frostburn: Let's defer to a later issue. I need to sit down to solve it so I cannot give a concrete solution while just glancing over.

src/components/ScaleLattice.vue
diegovdc: I don't think it makes sense to predefine all coordinates. In my opinion, a UI so that a user can add, remove and change coords for the primes they need would be more useful.

frostburn: Sure, we can do tiered issues. One for completing the predefined coords and another for adding user-configurability. Just let me know if you're strongly against predefinition beyond the 23-limit. To me, going up to 97-limit is an easy win and better than nothing.

src/views/LatticeView.vue
```diff
@@ -0,0 +1,40 @@
...
+  baseMidiNote: number;
+  keyColors: string[];
+  heldNotes: Map<number, number>;
+}>();
+
+const heldScaleDegrees = computed(() => {
+  const result: Set<number> = new Set();
+  for (const tableIndex of props.heldNotes.keys()) {
+    if (props.heldNotes.get(tableIndex)! > 0) {
+      result.add(mmod(props.baseMidiNote + tableIndex, props.scale.size));
```
frostburn, line 18: This should be `tableIndex - props.baseMidiNote`. Sorry it took so long to figure out.

src/utils.ts
```diff
@@ -238,3 +238,33 @@ export function computedAndError<T>(
   const error = computed(() => valueAndError.value[1]);
   return [value, error];
 }
+
+export function getPairCombinations<T>(arr: T[]): [T, T][] {
```
frostburn, line 242: Isn't this just `kCombinations(arr, 2)`?

src/utils.ts
```diff
@@ -238,3 +238,33 @@ export function computedAndError<T>(
...
+    for (let j = i + 1; j < arr.length; j++) {
+      combinations.push([arr[i], arr[j]]);
+    }
+  }
+
+  return combinations;
+}
+
+// Calculates euclidean distance for monzos, as it assumes that if points are not of the same dimension, then the missing dimensions have a value of zero.
+export function monzoEuclideanDistance(
```
frostburn, line 255: For future reference, this is just `norm(sub(point1, point2))` if we ever get around updating `xen-dev-utils` dependency. For efficiency even taxicab norm suffices in this application.

src/components/ScaleLattice.vue
```diff
@@ -0,0 +1,338 @@
...
+  try {
+    if (!Array.isArray(nodes.value)) {
+      return [];
+    }
+    const combinations = getPairCombinations(props.scale.intervals);
+    return combinations
+      .map(([a, b]) => {
+        return {
+          distance: monzoEuclideanDistance(
+            a.monzo.toIntegerMonzo().slice(1),
```
frostburn, line 197: Was the lattice supposed to work with all prime equaves? This line assumes an octave.

src/components/ScaleLattice.vue
diegovdc: Hmm I really don't have the time to figure out a good set of coordinates, I believe it requires much experimentation and a good sense of which sort of scales are actually going to be made, otherwise it's just randomness. 

Also as you can see here, the coordinate system is already quite cluttered and we only have 9 primes, adding 15 more seems like it'd be pretty hard to understand and distinguish between one or the other.
image


But if we wanted to make a default set, I believe that a community effort would be a better fit, if possible.  Perhaps opening a call for suggestions and then a round of votes by members of the community would  get us somewhere.

In that case adding user-configurability first would allow them to easily explore and propose something.

src/components/ScaleLattice.vue
```diff
@@ -0,0 +1,338 @@
...
+    minY: Infinity,
+    maxX: -Infinity,
+    maxY: -Infinity,
+  };
+  if (svgElement.value) {
+    return [
+      ...svgElement.value.querySelectorAll("text"),
+      ...svgElement.value.querySelectorAll("circle"),
+    ].reduce(({ minX, maxX, minY, maxY }, el) => {
+      const { x, y, width, height } = el.getBBox();
```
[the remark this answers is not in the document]
diegovdc: How do I do that?

src/utils.ts
diegovdc: One would also need to remove the equave index right? Otherwise the octave reduction could increase the distance.

frostburn: Yes, but you don't have to figure it out in this PR. Let's focus on a minimum viable product and defer everything we've talked about on Discord to a separate issue.

I suggest that you simply display an error message to the user if the equave isn't 2/1.

frostburn:
The scale [2/1, 3/1] (the octave with a tritave equave) is broken.
image

Simply drop the support for support for non-octave equaves and show an error message.

frostburn:
The scale [9/8, 2/1] is disconnected. It would be nice to see an edge split into two with a small tick mark here.
image

src/views/LatticeView.vue
```diff
@@ -0,0 +1,40 @@
+<script setup lang="ts">
+import { computed, ref } from "vue";
```
frostburn, line 2: Always remember to `npm run lint` before pushing.

src/components/ScaleLattice.vue
```diff
@@ -0,0 +1,351 @@
...
+const viewBox = computed(() => {
+  if (svgElement.value) {
+    return [
+      ...svgElement.value.querySelectorAll("text"),
+      ...svgElement.value.querySelectorAll("circle"),
+    ].reduce(
+      ({ minX, maxX, minY, maxY }, el) => {
+        const { x, y, width, height } = el.getBBox();
+        const strokeWidth_ = el.getAttribute("stroke-width") ?? "0";
+        const strokeWidth = parseFloat(strokeWidth_);
```
frostburn, line 219: Not what I meant. `el.getBBox` supports strokes if you tell it to.

frostburn:
Added deferred stuff to #38 but I highly recommend you already deal with the item:
*Refactor the code so that node&edge creation is a separate unit-tested utility*

The lattice component should take a pre-baked list of nodes, edges and labels and the code that produces such a graph should have a battery of unit tests to verify it's doing the right thing.

frostburn:
Deployed to https://scaleworkshop.lumipakkanen.com/ for user testing/feedback.

src/components/ScaleLattice.vue
```diff
@@ -0,0 +1,351 @@
...
+    },
+    {
+      x: 20,
+      y: -6,
+    },
+  ];
+
+const container = ref<HTMLDivElement | null>(null);
+
+// TODO improve
```
frostburn, line 66: Prefer not to leave comments in production code. If left in, prefer to reference the issue that explains what needs to be done like #38.

diegovdc:
> The scale [9/8, 2/1] is disconnected. It would be nice to see an edge split into two with a small tick mark here. image

can you describe a more generalized behavior for this?  In which cases shall this be shown.

frostburn:
> > The scale [9/8, 2/1] is disconnected. It would be nice to see an edge split into two with a small tick mark here. image
> 
> can you describe a more generalized behavior for this? In which cases shall this be shown.

When a node has no edges based on your original algorithm, it should connect to the closest node(s) along prime(s). That way there are no floating nodes and no confusing off-prime edges either.

diegovdc:
I'll have to think more about that, can we deal with that on another issue?

diegovdc:
> Added deferred stuff to #38 but I highly recommend you already deal with the item: _Refactor the code so that node&edge creation is a separate unit-tested utility_
> 
> The lattice component should take a pre-baked list of nodes, edges and labels and the code that produces such a graph should have a battery of unit tests to verify it's doing the right thing.

Sure, can do on the next PR

src/components/ScaleLattice.vue
```diff
@@ -0,0 +1,346 @@
...
+}
+
+const equavePrimeIndex = computed(() =>
+  maybeGetPrimeEquave(props.scale.equave)
+);
+
+const nodes: ComputedRef<Node[] | string> = computed(() => {
+  try {
+    if (equavePrimeIndex.value !== 0) {
+      return "The lattice currently only supports scales with a 2/1 equave";
```
frostburn, line 143: This doesn't seem to trigger. I still get a messy visual for the scale [2/1, 3/1].

frostburn:
I think this is good to go once there's a working error message for non-octave scales.
@SeanArchibald do you want to take a look before we merge to `main`?

diegovdc:
@frostburn are you sure you are on the latest commit?
I see this on my local:
image

frostburn:
> @frostburn are you sure you are on the latest commit? I see this on my local: image

Good call. It seems you need to do `git fetch` while on the branch if it's not from origin.

src/components/ScaleLattice.vue
```diff
@@ -0,0 +1,346 @@
...
+    </div>
+    <div v-else>
+      <svg
+        v-if="Array.isArray(nodes)"
+        ref="svgElement"
+        id="lattice"
+        class="lattice"
+        xmlns="http://www.w3.org/2000/svg"
+        preserveAspectRatio="xMidYMid meet"
+        style="background-color: transparent"
```
frostburn, line 274: Styles should be in the `<style scoped>` section.

src/components/ScaleLattice.vue
```diff
@@ -0,0 +1,346 @@
...
+          :y1="nodesByLabel[edge.pair[0]].y"
+          :y2="nodesByLabel[edge.pair[1]].y"
+        />
+        <circle
+          v-for="(node, index) of nodes"
+          :key="index"
+          :cx="node.x"
+          :cy="node.y"
+          r="2.5"
+          :class="`node   ${
```
frostburn, line 299: Vue supports dictionaries as classes so this reads better as something like `{node: true, heldDegreee: heldScaleDegrees.has(node.index)}`.

frostburn:
No functional issues found. Good to go once the minor style issues have been resolved.

Re-deploying to my mirror for testing...

frostburn:
Took care of the remaining issues in #459 .

frostburn:
Closing as obsolete.

Issue #461, opened 2023-11-02
Historical Temperament Menu Option for Meantone etc.


000masa000:
Meantone Temperament:
Generator is 3/2 less a fraction of a comma, which is generally one of 
Syntonic (quintal) : 81/80
Pythagorean (tertial): 531441/524288
Schisma (tertial - quintal): 32805/32768
but could also be a custom value, possibly a difference of two commas,
 i.e. Septimal Comma: 64/63 
septimal + tertial: 59049/57344
.....

Result generally is 12 note scale, although "extended" meantone with split keys is possible

Default from C is 3 downward fifths (to Eb) and 8 upward (to G#).

frostburn:
This sounds more like a preset than a genuine parametrizized scale generator.
Could you describe the algorithm how historical temperaments are built up so that we can generalize?
One the general modal has been implemented it can have some preset buttons for easy access to classic scales.

000masa000:
The general model is based on tuning 3/2 and 4/3, 5/4 and 5/3 in the middle register, and adjusting the intervals so they "beat similarly". The resulting scales are described as narrowing the fifths by fractions of a comma, which is almost always Syntonic or Pythagorean. However, since the original tunings were achieved by ear, establishing beating rates is more likely to be a criterion than exact logarithmic equality of deviation. 

I could imagine for a musician familiar with this approach that it would be useful to specify the narrowing of the 3/2's in fractions of a comma. Optimising the tuning could be calculated by comparing beating rates between partial unisons of near 3/2, 4/3, 5/4 and 5/3 relationships, and optimisation choices could be made. For example, in 1/5 Syntonic Comma meantone, the fifths are narrowed by 1/5 Syntonic Comma, so four fifths approximates 5/4 1/5 comma too big. Then the triad 4:5:6 could be tuned so the beating rates of the 5/4, 6/5 and 3/2 are correlated. For example with a temperament of 1/4.598701962 syntonic comma, the triads 4:5:6 have a beating ratio of 1:1 between the perfect fifth and major third....

frostburn:
Tuning stuff by ear doesn't sound like something that lends itself to a well-defined scale generator. Do you have a mathematical definition for "beating similarly". The last time I looked into it there were multiple valid ways to define it from plain frequency differences to effective envelopes which unfortunately depend on the timbre.

[...]

PR #466, opened 2023-11-07
Generate historical temperaments and target-tempered generator-stacks

frostburn:
Move the other Rank-2 temperament modal further down the list.

ref #461

### DIFF
--- a/src/assets/base.css
+++ b/src/assets/base.css
@@ -175,6 +175,10 @@ button:disabled:hover {
   color: var(--color-accent-mute);
   background-color: var(--color-accent-background);
 }
+input:disabled {
+  color: var(--color-accent-mute);
+  background-color: var(--color-background-mute);
+}
 ul.btn-group,
 .btn-dropdown-group ul {
   list-style: none;
--- a/src/components/ScaleBuilder.vue
+++ b/src/components/ScaleBuilder.vue
@@ -18,6 +18,7 @@ import ApproximateByHarmonicsModal from '@/components/modals/modification/Approx
 import SubharmonicSeriesModal from '@/components/modals/generation/SubharmonicSeries.vue'
 import EnumerateChordModal from '@/components/modals/generation/EnumerateChord.vue'
 import CpsModal from '@/components/modals/generation/CombinationProductSet.vue'
+import HistoricalModal from '@/components/modals/generation/HistoricalScale.vue'
 import CrossPolytopeModal from '@/components/modals/generation/CrossPolytope.vue'
 import LatticeModal from '@/components/modals/generation/SpanLattice.vue'
 import EulerGenusModal from '@/components/modals/generation/EulerGenus.vue'
@@ -162,6 +163,7 @@ const showEulerGenusModal = ref(false)
 const showDwarfModal = ref(false)
 const showRankTwoModal = ref(false)
 const showCrossPolytopeModal = ref(false)
+const showHistoricalModal = ref(false)
 
 const showRotateModal = ref(false)
 const showSubsetModal = ref(false)
@@ -250,6 +252,7 @@ function updateScaleAndHideModals(scale: Scale) {
   showDwarfModal.value = false
   showCrossPolytopeModal.value = false
   showLatticeModal.value = false
+  showHistoricalModal.value = false
   showRotateModal.value = false
   showSubsetModal.value = false
   showStretchModal.value = false
@@ -291,6 +294,7 @@ function confirmPreset() {
             <a href="#" @click="showEnumerateChordModal = true"><li>Enumerate chord</li></a>
             <a href="#" @click="showCpsModal = true"><li>Combination product set</li></a>
             <a href="#" @click="showMosModal = true"><li>Moment of symmetry scale</li></a>
+            <a href="#" @click="showHistoricalModal = true"><li>Historical temperament</li></a>
             <a href="#" @click="showEulerGenusModal = true"><li>Euler-Fokker genus</li></a>
             <a href="#" @click="showDwarfModal = true"><li>Dwarf scale</li></a>
             <a href="#" @click="showCrossPolytopeModal = true"><li>Cross-polytope</li></a>
@@ -538,6 +542,17 @@ function confirmPreset() {
       @update:scale="updateScaleAndHideModals"
       @update:keyColors="emit('update:keyColors', $event)"
       @cancel="showRankTwoModal = false"
+    />
+
+    <HistoricalModal
+      :show="showHistoricalModal"
+      :centsFractionDigits="centsFractionDigits"
+      @update:scaleName="emit('update:scaleName', $event)"
+      @update:scale="updateScaleAndHideModals"
+      @update:baseFrequency="emit('update:baseFrequency', $event)"
+      @update:baseMidiNote="emit('update:baseMidiNote', $event)"
+      @update:keyColors="emit('update:keyColors', $event)"
+      @cancel="showHistoricalModal = false"
     />
 
     <ShareUrlModal
--- a/src/components/modals/generation/HistoricalScale.vue
+++ b/src/components/modals/generation/HistoricalScale.vue
@@ -1,0 +1,869 @@
+<script setup lang="ts">
+import { ExtendedMonzo, Interval, parseLine, Scale } from 'scale-workshop-core'
+import Modal from '@/components/ModalDialog.vue'
+import { computed, reactive, ref, watch } from 'vue'
+import { DEFAULT_NUMBER_OF_COMPONENTS, FIFTH, OCTAVE } from '@/constants'
+import ScaleLineInput from '@/components/ScaleLineInput.vue'
+import { mmod, centsToValue, lcm, Fraction } from 'xen-dev-utils'
+import { mosSizes } from 'moment-of-symmetry'
+import { circleDifference, spineLabel as spineLabel_, type AccidentalStyle } from '@/utils'
+
+const props = defineProps<{
+  centsFractionDigits: number
+}>()
+
+const emit = defineEmits([
+  'update:scaleName',
+  'update:scale',
+  'update:keyColors',
+  'update:baseFrequency',
+  'update:baseMidiNote',
+  'cancel'
+])
+
+// The presets default to middle C when there's an 'F' in the scale i.e. down >= 1
+const KEY_COLORS_C = [
+  'white',
+  'black',
+  'white',
+  'black',
+  'white',
+  'white',
+  'black',
+  'white',
+  'black',
+  'white',
+  'black',
+  'white'
+]
+
+const MIDI_NOTE_C = 60
+const FREQUENCY_C = 261.6255653005986
+
+const MAX_SIZE = 99
+const MAX_LENGTH = 10
+
+const HARMONIC_SEVENTH = new Interval(
+  ExtendedMonzo.fromFraction('7/4', DEFAULT_NUMBER_OF_COMPONENTS),
+  'ratio'
+)
+
+const SYNTONIC = new Interval(
+  ExtendedMonzo.fromFraction('81/80', DEFAULT_NUMBER_OF_COMPONENTS),
+  'ratio'
+)
+
+const ACCIDENTAL_STYLE =
+  (localStorage.getItem('accidentalPreference') as AccidentalStyle) ?? 'double'
+
+function spineLabel(up: number) {
+  return spineLabel_(up, ACCIDENTAL_STYLE)
+}
+
+const method = ref<'simple' | 'target' | 'well temperament'>('simple')
+
+const generator = ref(FIFTH)
+const generatorString = ref('3/2')
+
+const size = ref(12)
+const down = ref(3)
+const format = ref<'cents' | 'default'>('cents')
+
+const selectedPreset = ref('pythagorean')
+
+const pureGenerator = ref(FIFTH)
+const pureGeneratorString = ref('3/2')
+const target = ref(parseLine('7/4', DEFAULT_NUMBER_OF_COMPONENTS))
+const targetString = ref('7/4')
+const searchRange = ref(11)
+const period = ref(OCTAVE)
+const periodString = ref('2/1')
+const pureExponent = ref(10)
+const temperingStrength = ref(1)
+
+const wellComma = ref(parseLine('531441/524288', DEFAULT_NUMBER_OF_COMPONENTS))
+const wellCommaString = ref('531441/524288')
+const wellCommaFractionStrings = reactive<Map<number, string>>(
+  new Map([
+    [-6, '0'],
+    [-5, '0'],
+    [-4, '0'],
+    [-3, '0'],
+    [-2, '0'],
+    [-1, '-1/6'],
+    [0, '-1/6'],
+    [1, '-1/6'],
+    [2, '-1/6'],
+    [3, '-1/6'],
+    [4, '-1/6']
+  ])
+)
+
+const selectedWellPreset = ref('vallotti')
+
+const mosSizeList = computed(() => {
+  const p = method.value === 'simple' ? 1200 : period.value.totalCents()
+  const g = temperedGenerator.value.totalCents()
+  const sizes = mosSizes(g / p, MAX_SIZE, MAX_LENGTH + 2)
+  if (p > 600) {
+    while (sizes.length && sizes[0] < 4) {
+      sizes.shift()
+    }
+  }
+  while (sizes.length > MAX_LENGTH) {
+    sizes.pop()
+  }
+  return sizes
+})
+
+type Candidate = {
+  exponent: number
+  tempering: number
+}
+
+const candidates = computed(() => {
+  const pureCents = pureGenerator.value.totalCents()
+  const targetCents = target.value.totalCents()
+  const periodCents = period.value.totalCents()
+  const halfPeriod = 0.5 * periodCents
+  const result: Candidate[] = []
+  for (let i = -searchRange.value; i <= searchRange.value; ++i) {
+    if (i === 0 || i === 1 || i === -1) {
+      continue
+    }
+    const offset = mmod(targetCents - pureCents * i + halfPeriod, periodCents) - halfPeriod
+    result.push({
+      exponent: i,
+      tempering: offset / i
+    })
+  }
+  result.sort((a, b) => Math.abs(a.tempering) - Math.abs(b.tempering))
+  return result
+})
+
+// This way makes the selection behave slightly nicer when other values change
+const tempering = computed(() => {
+  for (const candidate of candidates.value) {
+    if (candidate.exponent === pureExponent.value) {
+      return candidate.tempering
+    }
+  }
+  return 0
+})
+
+const strengthSlider = computed({
+  get: () => temperingStrength.value,
+  set(newValue: number) {
+    // There's something wrong with how input ranges are handled.
+    if (typeof newValue !== 'number') {
+      newValue = parseFloat(newValue)
+    }
+    if (!isNaN(newValue)) {
+      temperingStrength.value = newValue
+    }
+  }
+})
+
+const temperedGenerator = computed(() => {
+  const lineOptions = { centsFractionDigits: props.centsFractionDigits }
+  if (method.value === 'simple') {
+    return generator.value.mergeOptions(lineOptions)
+  }
+  return pureGenerator.value
+    .mergeOptions(lineOptions)
+    .add(
+      new Interval(
+        ExtendedMonzo.fromCents(
+          tempering.value * temperingStrength.value,
+          DEFAULT_NUMBER_OF_COMPONENTS
+        ),
+        'cents',
+        undefined,
+        lineOptions
+      )
+    )
+    .asType('any')
+})
+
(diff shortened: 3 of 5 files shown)


frostburn:
@000masa000 please review.

000masa000:
Hi I am looking at the historical temperaments on https://scaleworkshop.lumipakkanen.com/
Instead of 1/11 comma meantone (rare) you might instead do 1/8 Pythagorean comma, which is used by musicians. Generator `3/2-1\8<531441/524288>` 1/6 comma the generator should be 1/6 Pythagorean comma, so `3/2-1\6<531441/524288>`..... the 81/80 is used for 1/5 comma 1/4 comma 2/7 comma and 1/3 comma,

000masa000:
Number of fourths (down) I would set for 2 flats for Meantone (so a default value of 3, getting us to C-F-Bb-Eb) instead of 4 .
For Pythagorean tuning I would go from Db-F#, so a default value of 5 so the ratios from C are simplest for each chromatic

000masa000:
in the "target" tab a few typos: "Stacking interval" ("c" missing) "Search range" ("r" missing)
The feature is really cool! Needs a moment to figure out that the list tells you how much tempering there is per fifth based on a target, but once I understood the results are very powerful

frostburn:
Fixed found issues. Redeployed to https://scaleworkshop.lumipakkanen.com/

000masa000:
in the target mode of historical change default to 1 down, then C-Bb and F-Eb are 7/4s and the scale has a 7/4 and a 7/6 (nice)! looking good otherwise on a first glance

frostburn:
> in the target mode of historical change default to 1 down, then C-Bb and F-Eb are 7/4s and the scale has a 7/4 and a 7/6 (nice)! looking good otherwise on a first glance

Updated the code but will redeploy only tomorrow. I will also prompt XA for comments if there are no more code changes by then.

000masa000:
can commas be added using the syntax + and -? in that case I can make all the main Well Temperaments as presets, even though algorithmic generation is tricky but possible: organise the chain of 16 central 3/2's with 6 down (from C, so including all 5 flats)  and 10 up, the pitch range should run from Gb-Db-Ab-Eb-Bb-F-C-G-D-A-E-B-F#-C#-G#-D#-A# (as default) as a list of monzos with increasing powers of 3. Then a subset of 12 notes (11 ratios) will each be modified according to the following rule:

the total modification by the end of the chain (12th ratio, determined automatically) should be equal to one pythagorean comma 531441/524288. Along the way modifications are specified as fractions of either this comma or of a syntonic comma, or (in some cases) a sum of two commas (Pythagorean, Schisma). In the background, keep track of fractions of a schisma "left over".

If this sounds too confusing let me show a worst case example, the French Well-Temperament by Rameau (1726) tunes the following chain in 1/4 comma meantone : Bb through B, it leaves B-F#-C# as 3/2s, reduced C#-G# by 1/6 Pythagorean Comma, and then splits the difference. Thus, in the chain Bb-D# (2 down) we generate the following 12 ratios using the 1/4 syntonic method ---

 here it would be nicer (for my way of understanding the shorthand) for the scale display to express (optionally) the Pythagorean ratios and their modification rather than the Dirac bra-ket prime- or generator-count notation.... I am slightly confused by it, actually : can you clarify, see below, generated as 1/4 comma  2 down: the last number I surmise is the fraction of 81/80 used to temper the notes, which are Pythagorean, thus powers of 2 and 3: why are the first two numbers of the 3/2 0 and 0 instead of some powers of 2 and 3, i.e. -1 and 1 for example. Sorry for the question, but I don't fully understand the implementation....

1/4 comma with down 2:

[-4, 0, 7/4>
[-1, 0, 1/2>  this note is D 
[-5, 0, 9/4>
[-2, 0, 1>    
[1, 0, -1/4>  this note is F
[-3, 0, 3/2>  
[0, 0, 1/4>   this note is G `3/2-1\4<81/80>` (should not the prime count be "1")?
[-4, 0, 2>
[-1, 0, 3/4> this note is A
[2, 0, -1/2>
[-2, 0, 5/4>
2/1     this note is C



ANYWAY I would do it from fractions like this:

start with Pythagorean 2 down:

2187/2048 - `5\4<81/80>` ! (same deviation as B)
9/8 - `2\4<81/80>`
19683/16384
81/64 - `4\4<81/80>`  !( i.e. 5/4)
4/3 + `1\4<81/80>`
729/512 - `5\4<81/80>` ! (same deviation as B)
3/2 - `1\4<81/80>`
6561/4096 - `5\4<81/80> - 1\6<531441/524288>` ! additional 1/6 Pythagorean comma
27/16 - `3\4<81/80>`
16/9 + `2\4<81{80>`
243/128 - `5\4<81/80>` !end of 1/4 chain
2/1



at this stage there is one note to tune, D#, producing two fifths: G#-D# and D#-Bb. We have temepered by a total of 7/4 Syntonic + 1/6 Pythagorean. Measured in terms of the Pythagorean Comma, which is what we really want to temper out, we can rewrite this as 7/4(Pythagorean - Schisma) + 1/6(Pythagorean) = 23/12 Pythagorean - 7/4 Schisma. So the amount of comma we need to recouperate is 11/12 Pythagorean - 7/4 Schisma. If we divide the two remaining fifths equally, the amount of tempering applied to D# should be the same value as G# (producing a pure fifth) PLUS 1/2 of the remaining amount, (11/24 of a Pythagorean less 7/8 schisma) Calculating this by hand is mind bending, but the machine can do it easily with a button (divide remaining difference from 1 Pythagorean comma equally amongst the remaining fifths). In fact, this could be a cool way of expressing 12edo: it begins as a user specified Pythagorean tuning, modified by 1\12 Pythagorean. Then, as commas are modified manually, there could be an option to divide remaining commas in the chain equally. So if you set one fifth pure, the reamining ones can be adjusted accordingly. THis could of course be done in either direction (up the fifths or down the fourths)

000masa000:
one final thought: when pressing a preset button, it would be nice in the front end to see the selection has been made active (color change or radio button)  thanks!

frostburn:
> [0, 0, 1/4> this note is G 3/2-1\4<81/80> (should not the prime count be "1")?

The monzo of `81/80` is `[-4, 4, -1>` if you scale that by `-1\4` you get `[1, -1, 1/4>`.
Now add that to `3/2` i.e. `[-1, 1, 0>` and you get your G. I hope this clarifies why the monzos are like that.

frostburn:
That well-tempering explanation makes my head spin. Could you write out the algorithm in code instead of English? Does that algorithm generalize to generators other than `3/2`? If not, I would just add a fixed copy of every notable historical scale that doesn't have a rank-2 structure under *Load preset scale*.

000masa000:
thanks for the clear monzo explanation.

000masa000:
Well Temperament maybe best I hack together some presets and as we work on it the procedure / algorithm will clarify itself. Explaining is a bit complicated. Basically a chain of 12 fifths needs to have a Pythagorean comma taken away, equally or unequally, across the set of pitches. The usual method involves fractions of a Pythagorean or Syntonic Comma. If using fractions of a Syntonic, then one needs to keep track of the difference Pythagorean - Syntonic = Schisma. Each of these can be expressed as a monzo so it is not hard to notate the tempering. What might be useful is, once the desired commas are positioned, to be able to divide the remaining portion of the tempering across the remaining fifths, since this is kind of the way it was done historically.

000masa000:
for example: temper 8 fifths by 1/4 Syntonic, have several pure fifths, one tempered by 1/6 Pythagorean comma, then figure out the remainder and divide equally.

frostburn:
If there is an algorithm for well-tempering, it should go to `core` first.
If there isn't, it should go to presets,
Either way I think well-tempering is out of scope for this PR. It can go in this modal, but later.

frostburn:
Migth need to tweak the beating equalizer a bit so that it focuses on the multigen when the generator itself is discordant. That way you could create equally beating scales with neutral thirds for example.
Also I'd like to at least see the MOS sizes somewhere even if we keep things simpler than the proper Rank-2 modal.

frostburn:
Tweaked the beating equalizer to be aware of multigenerators and displayed MOS sizes.
(For testing multigens remember that e.g. the neutral third is spelled `1\2<3/2>` in our syntax.)
Latest version deployed at: https://scaleworkshop.lumipakkanen.com/

frostburn:
@000masa000 Replaced the preset names with ASCII instead of fancy unicode vulgar fractions. Deployed the latest version to the beta server. Would you consider this a good first version of the historical modal dialog? We can always fine-tune things, implement non-western presets and add more radio-tabs in future pull requests.

000masa000:
I would remove 1/2 comma since it isn't particularly useful nor historically important. I am happy to suggest some more presets, one might be the tuning Helmholtz used (tempering the 3/2s so that -8 fifths = 5/4) with 24 pitches. I need to check how many fifths down his organ used. It would be nice to have a well-temperament section, but maybe I will try to explain it is a clearer way and we can see how it might work.

frostburn:
Removed 1/2 from rank-2 historicals. Added *Helmholtz aka Schismatic* with down of `11`.
Added a radio tab for Well Temperaments. Added the ones you provided images for as presets.
In my implementation the circle doesn't always close (that happens only if one whole pythagorean comma is being distributed) but I try to provide information about the enharmonics to compensate.
Deployed to the beta server for testing.

frostburn:
I just noticed that there's some :monkey_face: business with the name when changing radio tabs. Will fix later...

000masa000:
HI, checking Kellner: I notice that if the reference is set to A 440, it is automatically the first note of the scale (1/1) as well. Seems to me like there should be an option to set the starting note the C (which automatically rotates the scale to the "correct" position) I will open an issue if you like. If I set the starting note to the frequency of C and the MIDI to 60, Kellner comes out correct

000masa000:
adding some more well temperaments in discord

frostburn:
I think most if not all of these presets start at C. The default in SW has been A because that's the A440Hz pitch standard and the key colors follow. It's not a huge hassle to make this modal affect everything from base MIDI note to base frequency to key colors.

frostburn:
Added rest of the well temperaments.
Made all 12 note presets start from MIDI note 60 with the corresponding base frequency and key colors for middle C.
Fixed the name issues.
Redeployed to beta server.

frostburn:
@000masa000 Can you double check *Corrette*? The enharmonic distance is suspiciously large.

000masa000:
yes, that is correct, it is a comma too big.... better than 1/4 comma but still bad!

frostburn:
> yes, that is correct, it is a comma too big.... better than 1/4 comma but still bad!

Can you spot if I made an error or if there's an error in the source image?

000masa000:
Vallotti 1754
Neidhardt Kleine Stadt 1732
Neidhardt Große Stadt 1724
Bach/Barnes 1979
Bach/Lehman 2005
Bach/O'Donnell 2006    ----(note: 2 ll's) 
Bach/Hill 2008
Bach/Swich 2011
Bach/Louie 2018
Lambert 1774
Grammateus 1518
Kellner 1975
Werkmeister III 1691
Kirnberger III 1779
Corrette 1753
Marpourg 1756
Egarr's English Ord 2007
Rameau 1726

000masa000:
and here is one more:

d'Alembert/Rousseau 1752

image

frostburn:
Any recommendations for the order of the presets? I'm doing pythagorean comma first and syntonic comma second with random order within. Should they be ordered by date within?

000masa000:
by date is nice

frostburn:
There's this quote about *Corrette*:

> Corrette began his temperament from F, tuning eight Quarter-comma narrow fifths in a row, followed by a less-narrow fifth. Two other fifths are wide, and as always, the remaining G♯– E♭ wolf is not intentionally tuned: Corrette called it the ”deffaut de la Partition”.
from https://www.hpschd.nu/index.html?nav/nav-4.html&t/welcome.html&https://www.hpschd.nu/tech/tmp/corrette.html

So I would guess that the wolfy enharmonic is intentional. I can't find a mistake in my work compared to the image.

000masa000:
sounds right.... so Corrette is not really a well temperament, maybe could just leave it out.

frostburn:
I'm going to rename it to *Corrette's Wolf* to distinguish it from the rest. It still follows the same pattern without requiring extra configuration so I vote for it to stay.

000masa000:
I have made two rational well temperaments, which would like rationalisations-by-hand of vallotti and a 1/3 comma kirnberger variant.... here they are as potential presets, similar sound, different structure, maybe could be included if there was either a way to indicate the deviation between fifths in cents, or as ratios, or if there was an option to approximate by ratios in the menu... ? what do you think?

000masa000:
Sabat I
12
!
17/16 ! C♯=17°/C
9/8 ! D
81/68 ! E♭=u17\E
81/64 ! E
4/3 ! F
27/19 ! F♯=u19\A
3/2 ! G
1731/1088 ! G♯|A♭
27/16 ! A
57/32 ! B♭=19°/G
243/128 ! B
2/1 ! C


Sabat II
12
!
16/15 ! ^D♭
9/8 ! D
81/68 ! E♭=u17\E
81/64 ! E
4/3 ! F
729/512 ! F♯
3/2 ! G
51/32 ! G♯=17°/G
27/16 ! A
3645/2048 ! vA♯
243/128 ! B
2/1

frostburn:
> possible bug, tried to implement well temperament with 31 notes and 15 down, alternating -1/6 and -1/3 of 81/80 symmetrically around C. The message a the end was off by a diesis, it did not calculate the right final fifth in the bigger circle.... Should say difference between D## and Fbb

I wasn't sure how to classify enharmonics as all of the historical examples use 12 notes. I went with the definition that enharmonics are separated by 12 steps. I'll switch to using the (start point + 0) and (end point + 1) and trust that the user knows what they're doing if they're cramming more than 12 notes into the spiral.

000masa000:
OK, the scale I am getting seems to produce a wrong cents value on the 30th note, for some reason it starts with 0. which it shouldn't.... and 1241 is one diesis over the octave....

0.
77.84119
115.315667
152.097999
193.156857
234.215715
270.998047
308.472524
349.531382
386.313714
427.372572
464.154904
501.629381
542.688239
579.470571
620.529429
657.311761
698.370619
735.845096
772.627428
813.686286
850.468618
891.527476
929.001953
965.784285
1006.843143
1047.902001
1084.684333
1122.15881
1241.058858
1200.

frostburn:
> I have made two rational well temperaments, which would like rationalisations-by-hand of vallotti and a 1/3 comma kirnberger variant.... here they are as potential presets, similar sound, different structure, maybe could be included if there was either a way to indicate the deviation between fifths in cents, or as ratios, or if there was an option to approximate by ratios in the menu... ? what do you think?

I feel like it complicates the process too much. All of the historical examples follow the recipe of distributing a single comma along the spine. We should stick to that for now.

There could be a power-user version in the future that supports generators other than 3/2, arbitrary periods and custom offsets per generation.

000masa000:
image
alternating -1/6 and -1/3, with two -1/6ths around C.... should be similar to 1/4 comma meantone extended to 31 notes..... hm

000masa000:
IMG_9925 11899

frostburn:
> OK, the scale I am getting seems to produce a wrong cents value on the 30th note, for some reason it starts with 0. which it shouldn't.... and 1241 is one diesis over the octave....

Not an easy issue to debug as it's related to how the generated intervals are calculated and reduced. Needs a battery of unit tests. I'm inclined to moving this issue to the `core` and freezing development on this PR until we have a well-tested scale generator for well temperaments.

000masa000:
OK, too bad, I really like it! I guess it is just grabbing an extra fifth in the wrong direction and ending up on the wrong side of the octave.....  .........................................Marc Sabatmobile DE +49 176 24569023mobile FI +358 ***@***.***://www.plainsound.org From: Lumi PakkanenSent: Thursday, 16 November 2023 18:42To: xenharmonic-devs/scale-workshopCc: Marc Sabat; MentionSubject: Re: [xenharmonic-devs/scale-workshop] Generate historical temperaments and target-tempered generator-stacks (PR #466) OK, the scale I am getting seems to produce a wrong cents value on the 30th note, for some reason it starts with 0. which it shouldn't.... and 1241 is one diesis over the octave....Not an easy issue to debug as it's related to how the generated intervals are calculated and reduced. Needs a battery of unit tests. I'm inclined to moving this issue to the core and freezing development on this PR until we have a well-tested scale generator for well temperaments.—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you were mentioned.Message ID: ***@***.***>

frostburn:
Please try to reply directly in GitHub. That email boilerplate is horrific.

000masa000:
I seem to have found the bug. If you choose more than 12 fifths, and set down to 0, then B# is higher than C by a Pythagorean comma. If you set C G to -2 syntonics, B# should be less than C, but instead it generates the scale degree 0. Probably needs to wrap around somehow....

frostburn:
Hopefully fixed the bug with your scale. (I made a 19 note scale while testing.)
Protip: You can specify generator offsets in cents by entering `1.0` as the comma for well tempering. The offset syntax still needs to be fractional so `1/10` instead of `0.1`. This is just a hack. I'll make a separate issue for the power-user version.

frostburn:
Fixed the size reseting issue. Associated the comma fractions with the spine.

000masa000:
Would it be possible in the menu for designing the scale to have a reset to 0 button for the 3/2 chain deviations, and a field for the scale name?

000masa000:
suggestion for triple flats, to use the glyph ♭ three times, because the double-flat and single flat combo looks awkward. In fact I prefer ♭♭♭ , ♭♭, ♭ and ♯, ♯♯, ♯♯♯ rather than the cryptic and confusing "double sharp" sign that no one can really read :)

frostburn:
> Would it be possible in the menu for designing the scale to have a reset to 0 button for the 3/2 chain deviations, and a field for the scale name?

Yes to the first one.

No to the second one. None of the other generation models have this kind of clutter and you can easily edit the name after you're done with the well tempering phase.

frostburn:
> suggestion for triple flats, to use the glyph ♭ three times, because the double-flat and single flat combo looks awkward. In fact I prefer ♭♭♭ , ♭♭, ♭ and ♯, ♯♯, ♯♯♯ rather than the cryptic and confusing "double sharp" sign that no one can really read :)

I'm of the opposite preference. I'll turn this into a user preference with options to use doubles, singles or even ASCII.

frostburn:
Fixed issues I agree with. Redeployed to https://scaleworkshop.lumipakkanen.com/

000masa000:
Working with the well temperament generator. If I mistype comma to an invalid value by mistake and leave the dialogue, when I return the entry field for comma (which has become an invalid /red value) disappears.... couldn't the comma be any value, positive or negative? Thanks for the accidentals option it is really nice to have the extended chain with various display options!

000masa000:
A thought about the reset deviations button, maybe (since it can be at the end of a long scroll list with hidden deviations), it would be nice to have up top; is there some unobtrusive way to show if there ARE non zero deviations in the list?

frostburn:
> A thought about the reset deviations button, maybe (since it can be at the end of a long scroll list with hidden deviations), it would be nice to have up top; is there some unobtrusive way to show if there ARE non zero deviations in the list?

It probably should be at both ends then. I can make it disabled if everything is zero already.

frostburn:
> Working with the well temperament generator. If I mistype comma to an invalid value by mistake and leave the dialogue, when I return the entry field for comma (which has become an invalid /red value) disappears.... couldn't the comma be any value, positive or negative? Thanks for the accidentals option it is really nice to have the extended chain with various display alternatives!

Well that's a new one! And not unique to this PR. You can do that :magic_wand: trick to any entry field that accepts intervals.

000masa000:
Hi! I am wondering if it is possible to not only have note names but an enumeration of the interval? It is useful for exploring where in a chain to set a target modification. I am experimenting with tweaking 53 3/2's towards rational targets, for example taking the "nearest 11/8" and seeing ways of hitting it nicely from the center pitch. This kind of stuff could be useful power tools, checking for fractional unequal distributions of a comma to smoothly approach multiple targets (like a move towards higher rank temperament....)

frostburn:
> Hi! I am wondering if it is possible to not only have note names but an enumeration of the interval? It is useful for exploring where in a chain to set a target modification. I am experimenting with tweaking 53 3/2's towards rational targets, for example taking the "nearest 11/8" and seeing ways of hitting it nicely from the center pitch. This kind of stuff could be useful power tools, checking for fractional unequal distributions of a comma to smoothly approach multiple targets (like a move towards higher rank temperament....)

That's a bit too much for a simple modal. Maybe we can have power tools in SW3 that take up the whole screen and display useful widgets while creating such complex scales with multiple optimization criteria.

Could you write the specs for the power-user version in a new issue? Feel free to include e.g. hand-drawn pictures to sketch out the UI.

frostburn:
Ready for review again.
Deployed to https://scaleworkshop.lumipakkanen.com/ .

frostburn:
Self-reviewing: Good enough. We can iterate on the design later.

Issue #471, opened 2023-11-12
Implement VAO / Concordance Shell generator


frostburn:
**Puhts** — _Today at 7:31 PM_

> I don't know much about the math implementation side of things, but the ability to calculate Vertically Aligned Objects (VAOs)/Concordance Shells in a desired ED within a specified error limit? 

> Yeah, I guess what I'm suggesting is implementing into SW so that way it's more accessible for those of us who are not code minded. Plus, making it more searchable rather than just a big list.

frostburn:
Likely relevant code written in Python / pandas: https://github.com/ZymoXyl/concordance_shells

frostburn:
Technically implemented. Couldn't find any resources on this so we'll improve if there's feedback.

PR #472, opened 2023-11-13
Implement algorithms for finding concordance shells

frostburn:
ref #471


frostburn:
@ZymoXyl please review.

frostburn:
The end result should probably be an octave-equivalent scale which means that everything needs to be reduced and the equivalence test needs to be reworked.

frostburn:
Decided to update the *Equalize* modifier while at it.
Will make the Concordance Shell modal soonish.

PR #481, opened 2023-11-18
Fix scale line input

frostburn:


frostburn:
I think I'll self-approve this one as welll so that everything can be deployed in one giant update.

PR #484, opened 2023-11-25
Update Scale Workshop ecosystem dependencies

frostburn:


frostburn:
Self-reviewing: Tests pass and I guess it builds. We shall test in production, now that we're running two in paralllel!

Issue #485, opened 2023-11-25
Equalize scale crashes with error "Invalid cents value"


frostburn:

frostburn:
To reproduce: enter `1/1` as the sole scale value and use the "Equalize" modifier with default settings.

PR #486, opened 2023-11-25
Update core dependency

frostburn:
Fixes unison equalization.

ref #485


frostburn:
Self-merging because it's the season.

Issue #411, opened 2023-03-25
Add a toggle for interval matrix indexing


frostburn:
**intars** (accidentally Latvian) — Today at 2:46 AM

> @ Sevish @ frostburn I wish the interval matrix in Scale Workshop used zero-indexing for degrees.
> Zero-indexing (I use "k-step" and "k-degree", where the "perfect 0-step/degree" is the unison/tonic) is just much better for referring to intervals and degrees in arbitrary scales.

**Scott Dakota** — Today at 3:14 AM

> A toggle switch for zero-index or one-index could be handy.

Issue #445, opened 2023-06-18
Add the option to hide the first column of the interval matrix


frostburn:
**inthar** — 06/16/2023 4:37 PM

> I don't know if I've already raised this issue, but the first column of the interval matrix is wholly unnecessary.

![image](https://github.com/xenharmonic-devs/scale-workshop/assets/1253499/171a69cf-c77e-4fa8-be3a-46257ec3bd20)

PR #500, opened 2023-11-27
 Add a toggle for interval matrix indexing (Issue #411)

inthar-raven:


inthar-raven:
Added radio buttons for "Interval indexing": 0-indexing (default) and 1-indexing.

frostburn:
1. Missing the actual `ref` that would be manipulated by the `v-model`
2. Commits should be squashed into one
3. The value should be persisted in local storage

frostburn:
The commit message should end with `ref #445`.

frostburn:
Or do we actually have duplicate issues? #411 is just fine then.

src/views/AnalysisView.vue
```diff
@@ -100,9 +101,9 @@ const matrix = computed(() => {
         <tr>
           <th></th>
           <th v-for="i of Math.min(scale.size, MAX_SCALE_SIZE)" :key="i">
-            {{ i }}
+            {{ i - 1 + indexing.value }}
           </th>
-          <th>({{ scale.size + 1 }})</th>
+          <th>({{ scale.size + indexing.value }})</th>
```
frostburn, line 106: In `<template>` code this is just called `indexing`. Now it shows NaNs.

Issue #367, opened 2022-11-29
Use Pinia to manage app state


frostburn:
This should reduce prop-drilling boiler-plate code quite a bit.

Also consider managing state for individual modals. Currently it's random if the modals remember previous configuration when trying different scales in succession.

PR #504, opened 2023-11-28
Use Pinia to manage state related to audio

frostburn:
ref #367


frostburn:
Self-reviewing: Seem fresh :pineapple: .

PR #505, opened 2023-11-29
Add a some scales found in the wild for verifying parser stability

frostburn:


frostburn:
Self-reviewing: Seems responsible to test using these.

Issue #508, opened 2023-11-30
Base frequency is not encoded in the URL


frostburn:
**inthar** — _Today at 7:52 PM_

> Base frequency resets to 440 Hz every time I change anything in the Synth view.

frostburn:
Decoding works though.

PR #509, opened 2023-11-30
Fix base frequency encoding in URL

frostburn:
Access a reactive property to trigger the watcher. Implement Cypress test to verify the fix.

ref #508


frostburn:

PR #519, opened 2023-12-30
Support primes as equaves in the lattice visualization

frostburn:


frostburn:
The direction of 2/1 keeps changing and causes confusion. Define new fixed coordinates that get zeroed out when the equave is an octave.

frostburn:
Made 2/1 a bit longer. Prime lattices with various equaves look reasonably good. Merging.

Issue #520, opened 2024-01-02
Break ScaleBuilder into async components


frostburn:
`dist/assets/index-[hash].js` is too large which is mostly caused by the index loading every modal at once.

Break the imports in `ScaleBuilder.vue` into [async components](https://vuejs.org/guide/components/async).

PR #521, opened 2024-01-02
Break ScaleBuilder into async components

frostburn:
ref #520


frostburn:
The main effect is to move all the math stuff related to tempering into a separate chunk that is not loaded on the front page making it over 2 megabytes lighter.

SeanArchibald:
The build produces a small index chunk. Also the modals worked as expected for me (I see there is already a PR for the Korg modal)

Issue #336, opened 2022-11-09
Split code into re-usable packages


frostburn:
- [x] Monzos, scales and parsing
- [x] MIDI
- [x] Typing keyboard
- [x] Synth

PR #523, opened 2024-01-03
Split isomorphic QWERTY code to a re-usable package

frostburn:
ref #336


frostburn:
I'll figure out something for the piano-style mappings, but this PR could be merged as is.

frostburn:
I did get the piano-style mappings to work, but there are some peculiarities with the existing approach that are hard to reproduce and preserve.

Colors and QWERTY/MIDI indexing will be redone in version 3 anyway, so I'll leave keyboard mapping as is for now.

SeanArchibald:
This worked as expected in my testing

PR #525, opened 2024-01-06
Close the Korg export modal when pressing Escape

frostburn:
ref #524


SeanArchibald:
Works well for me

frostburn:
It's a small enough change to merge without rebasing.

Issue #529, opened 2024-01-10
Store modal state in Pinia


frostburn:
**tristan bay** — _Today at 9:40 AM_

> I noticed that in the newest SW release, on mobile, when you finish creating a new scale using one of the tools, all of the settings are immediately cleared and what I used isn't there anymore when I open the same menu back up. This used to only happen when I would switch pages (e.g. to virtual keyboard) and then back. I liked having the scale settings still be there because if I didn't feel something was quite right (e.g. too many of a certain generator in a lattice), I could go back and quickly fix it without typing everything in again. Is this something you're willing to add back in, maybe even keeping settings after switching pages back and forth?

PR #535, opened 2024-01-16
Store modal state in Pinia

frostburn:
ref #529


frostburn:
You should be able to export tempering state as `{...state}` without spelling everything out twice.

frostburn:
Tested by couple of people so should be good. Merging.

Issue #553, opened 2024-02-10
Rank-2 modal error when changing number of periods


frostburn:
`RangeError: Invalid array length`

PR #555, opened 2024-02-12
Update xenharmonic-dev dependencies

frostburn:
ref #553


frostburn:
Self-reviewing: Seems OK and fixes the bug.

PR #621, opened 2024-03-25
Include windows-latest in continuous integration

frostburn:


frostburn:
The failing tests should be fixed in the scope of this PR.

Issue #617, opened 2024-03-24
Analysis tab: Add "Abstract dyads" option under "Display intervals in matrix as"


inthar-raven:
Most useful for scales with 4 or fewer step sizes, this feature would get all the step sizes in a scale and display each table entry as a linear combination of the step sizes.

inthar-raven:
For a scale of the form:
```js
let x = 7/6
let y = 8/7
let z = 9/8
x; y; x; y; z;
stack()
```
the interval matrix with that setting would look like:
```
     |  1  |  2  |  3  |  4   |  (5)  | 
---------------------------------------
0    |  x  | x+y |2x+y |2x+2y |2x+2y+z|
x    |  y  | x+y |x+2y |x+2y+z|2x+2y+z|
x+y  |  x  | x+y |x+y+z|2x+y+z|2x+2y+z|
2x+y |  y  | y+z |x+y+z|x+2y+z|2x+2y+z|
2x+2y|  z  | x+z |x+y+z|2x+y+z|2x+2y+z|
```

PR #619, opened 2024-03-25
Enter v3 beta cycle / The SW3 commit

frostburn:
**Warning!** Merging this PR will cause `main` to enter beta cycle where key features like URL sharing are disabled until a backend is implemented.
The state of `main` prior to merging will become `two` and continue to receive v2 bug fixes until it's deprecated at the end of v3 beta cycle.

This commit swaps out the core engine from scale-workshop-core to sonic-weave.
Too many breaking changes have been introduced to maintain an incremental commit history at this juncture.

PR #622, opened 2024-03-25
Added "Step vector" radio button under "Display intervals in matrix as" (issue #617)

inthar-raven:


src/analysis.ts
```diff
@@ -100,8 +100,9 @@ export function utonalFundamental(frequencies: number[], maxDivisor = 23) {
   return divisor * frequencies[0]
 }
 
-// Interval matrix a.k.a the modes of a scale
-export function intervalMatrix(scale: Scale) {
+// Interval matrix a.k.a the modes of a scale.
+export function intervalMatrix(scale: Scale): [[Interval]] {
```
frostburn, line 104: This is not the correct type syntax.
image
Please use an editor with a working TypeScript plugin.

src/analysis.ts
```diff
@@ -100,8 +100,9 @@ export function utonalFundamental(frequencies: number[], maxDivisor = 23) {
...
+  console.log('intervalMatrix was called')
```
frostburn, line 105: No `console.log` in PRs without marking them as drafts.

src/analysis.ts
```diff
@@ -110,3 +111,78 @@ export function intervalMatrix(scale: Scale) {
...
+  for (let i = 0; i < scale.size; ++i) {
+    const mode = scale.rotate(i)
+    result.push(columns.map((j) => getStepVector(mode, j)))
+  }
+  return result
+}
+
+// Get j-step vector on tonic of `scale` and return it as an array of [String, number] pairs
+// (sorted in descending order by size).
+export function getStepVector(scale: Scale, intervalClass: number): [[String, number]] {
```
frostburn, line 130: Re-calculating the step vectors for each cell in the matrix is a waste of compute.

frostburn:
My suggestion for the size indicators (smallest to largest).

1. x
2. s, L
3. s, m, L
4. s, m, M, L
5. s, m, n, M, L
6. s, S, m, M, l, L
7. s, S, m, n, M, l, L
8. s, S, m, n, N, M, l, L
9. a, b, c, d, e, f, g, h, i
10. (alphabet up to 10 first letters)
11. (etc.)

src/analysis.ts
```diff
@@ -110,3 +111,78 @@ export function intervalMatrix(scale: Scale) {
...
+  const scaleAsSteps = [...Array(scale.intervals.length).keys()].map((i) =>
+    scale.rotate(i).getInterval(1)
+  ) // the scale as steps
+  const sortedUniqueSteps = scaleAsSteps.filter(function (item, pos) {
+    // This function asks, "Is it the first `pos` such that `rep(item)` is that value?"
+    // (I had to do this because of the whole Proxy vs. Object weirdness.)
+    return scaleAsSteps.findIndex((step) => step.toString() === item.toString()) == pos
+  })
+  // Sort the scale steps
+  sortedUniqueSteps.sort().reverse() // sort in descending order
```
frostburn, line 143: `Array.sort` is a trap. You need to do `foo.sort((a, b) => a.compare(b))` or whatever I had implemented in `scale-workshop-core` if I had implemented anything...

src/analysis.ts
```diff
@@ -110,3 +111,78 @@ export function intervalMatrix(scale: Scale) {
...
+      keys = ['L', 'm', 's']
+      break
+    }
+    case 4: {
+      keys = ['L', 'm', 'n', 's']
+      break
+    }
+    default: {
+      keys = ['L'].concat(
+        [...Array(sortedUniqueSteps.length - 2).keys()].map((i) => 'm'.concat(i + 1)),
```
frostburn, line 167: Type error. Please don't abuse JS by concatenating integers to strings.

src/analysis.ts
inthar-raven: I think it's because I didn't import Interval. I should remove that declaration.

frostburn: Then it's a valid type, but still the wrong one. The return type seems to be `Interval[][]`.

arseniiv: That was syntax used in Haskell.

inthar-raven: I just use Linux Mint's default text editor. The npm didn't catch any type errors, and I should have asked why.

frostburn: `npm run type-check` does indeed pass. That's on Vue for writing a poor variant of TS.

src/analysis.ts
inthar-raven: Sorry, I forgot to check.

src/analysis.ts
inthar-raven: Yeah, I need to somehow do this "statically wrt the `Scale`".

frostburn:
This PR proves the concept but it comes at a bad time:

- SW2 shouldn't be receiving new features that might cause new bugs or performance issues before EOL and the Big Freeze
- The code relies on `scale-workshop-core` and I'd need to do a lot of error-prone rebasing/porting to `sonic-weave` if this is merged before #619

I suggest adding the basic functionality for manipulating and formatting abstract periodic strings to https://github.com/xenharmonic-devs/moment-of-symmetry
We can then utilize it in SW3 when the beta cycle begins.

inthar-raven:
> This PR proves the concept but it comes at a bad time:
> 
>     * SW2 shouldn't be receiving new features that might cause new bugs or performance issues before EOL and the Big Freeze
> 
>     * The code relies on `scale-workshop-core` and I'd need to do a lot of error-prone rebasing/porting to `sonic-weave` if this is merged before [Enter v3 beta cycle / The SW3 commit #619](https://github.com/xenharmonic-devs/scale-workshop/pull/619)
> 
> 
> I suggest adding the basic functionality for manipulating and formatting abstract periodic strings to https://github.com/xenharmonic-devs/moment-of-symmetry We can then utilize it in SW3 when the beta cycle begins.

I could first check if overhauling the core to `sonic-weave` actually breaks it.

inthar-raven:
> I suggest adding the basic functionality for manipulating and formatting abstract periodic strings to https://github.com/xenharmonic-devs/moment-of-symmetry We can then utilize it in SW3 when the beta cycle begins.

Yeah. This new word manipulation stuff could use a new module, anyway.

inthar-raven:
A couple more things we need to keep in mind when adapting this to SW3:
1. The variety for each interval class obviously needs to be at the step vector level in this mode
2. CS is never violated at the step vector level

frostburn:
> A couple more things we need to keep in mind when adapting this to SW3:
> 
> 1. The variety for each interval class obviously needs to be at the step vector level in this mode
> 2. CS is never violated at the step vector level

1. I think we should show abstract variety vs. actual variety
2. Nuh-uh: all x violates CS. (And seriously we should show actual CS violation even if abstractly there is none.)

inthar-raven:
>     2. Nuh-uh: all x violates CS. (And seriously we should show actual CS violation even if abstractly there is none.)
What I meant is that if a step vector has k steps, it obviously can never be another step vector with a different number of steps, though these two step vectors might be tuned the same. However, I must concede you have a point: CS violations at the tuning level would be useful to see.

Issue #286, opened 2022-09-01
Show note labels on virtual keys


frostburn:
- [ ] Scale lines
- [ ] Letters of the physical keyboard

PR #628, opened 2024-03-28
Make the key index number viewable

mpchadwick:
First time working with Vue here, so this is maybe not the best way to do this.

My problem is using the, virtual keyboard I'm confused which notes the keys are mapping to. I'm a new user of scale-workshop (just found it last weekend) and have been exploring different tonal system using the "load default" functionality.

Eventually (at least in my opinion) it would probably be helpful to have a setting that would overlay the key number and/of frequency over the keys to make it more clear for non-developers that wouldn't know how to find this info via DevTools.

For now I was just adding this because I find it useful at least. I can pop-open DevTools and double-check which key number for each key in the keyboard, and then cross reference that back to the build scale tab.

### DIFF
--- a/src/components/VirtualKeyboard.vue
+++ b/src/components/VirtualKeyboard.vue
@@ -57,6 +57,7 @@ const isMousePressed = ref(false)
           'hidden-sm': key.x > 8,
           held: (heldNotes.get(key.index) || 0) > 0
         }"
+        :index="key.index"
         :color="key.color"
         :isMousePressed="isMousePressed"
         :noteOn="() => noteOn(key.index)"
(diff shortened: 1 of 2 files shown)


frostburn:
Hi Max! We're in the middle of a beta cycle here. See the `two` branch if you need a working local copy. The `main` branch will be in shambles for couple of months.

The issue you raised would be #286 we just haven't gotten around to it yet.

mpchadwick:
I was able to run main branch the app locally actually without a problem and this change did do what I was looking for. It felt pretty harmless / non-intrusive, so I opened the PR as I thought someone else might find it useful, but feel free to close if you don't think it's relevant

frostburn:
Seems harmless enough. I'll review today at some point.

frostburn (approved):

frostburn:
Now live at: https://sw3.lumipakkanen.com/?version=3.0.0-beta.4

PR #629, opened 2024-03-29
Update sonic-weave dependency

frostburn:
Adds stepString() which labels the last interval in the scale with the step pattern.
Fix parsing if negative integers.
Make function parameters required by default.
Add array access with arrays of indices and of booleans.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "scale-workshop",
-  "version": "3.0.0-beta.2",
+  "version": "3.0.0-beta.3",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "scale-workshop",
-      "version": "3.0.0-beta.2",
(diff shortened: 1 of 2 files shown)


inthar-raven (commented):
This commit would just update `main` to beta 3, right?

frostburn:
> This commit would just update `main` to beta 3, right?

Correct.

inthar-raven:
Does the array access with indices work like `map array indices` (in pseudo-Haskell, viewing the array as a function)? And the Boolean thing is the array applied to the `true` indices, right?

frostburn:
> Does the array access with indices work like `map array indices` (in pseudo-Haskell, viewing the array as a function)? And the Boolean thing is the array applied to the `true` indices, right?

It's been a while since I've programmed in Haskell. I think what you're saying is true. The model I'm trying to emulate is numpy for Python.

inthar-raven (approved):

PR #630, opened 2024-03-31
Swap the meaning of 'a' and 'A' in absolute and relative notation

frostburn:
Update sonic-weave dependency.
Adapt to core syntax changes in presets and code generators.

Associated sonic-weave changelog:
Add Â as an alternative for augmented.
Convert NEDJI projection to a binary operation with 'ed' Add more tiers of operator precedence.
Make rounding operators bind looser than harmonic segments. Make min and max bind looser than addition.
Rename stdlib 'ed' to 'tet' to avoid the now-reserved word. Swap u for w for step variety 8 to 13.
Implement a single command to organize() the scale.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "scale-workshop",
-  "version": "3.0.0-beta.4",
+  "version": "3.0.0-beta.5",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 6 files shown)


frostburn:
Test build available at: https://sw3.lumipakkanen.com/?version=3.0.0-beta.5

000masa000 (approved):

PR #631, opened 2024-03-31
Bypass frequency range optimization with extreme scales

frostburn:
Format NaN and Infinity as is in the tuning table.


frostburn:
Straightforward enough. Merging due to lack of negative review.

Issue #640, opened 2024-04-08
SW3: Remove pythagorean enharmonic config


frostburn:
Just default to a sharp following the MIDI standard.
Power users know how to say `Eb4 = mtof(_) = 1/1`.

000masa000:
I wonder if the default Pythagoreans ought to be
A
Bb
B
C
C#
D
Eb
E
F
F#
G
G#
.... these are "closest ratios" from D = 1/1  (centred in the notation system)

frostburn:
MIDI is a not a great standard, but it is a standard nonetheless. It's what SW3 already uses for the labels of the empty scale that defaults to 12TET.

Bad defaults also incentivize people to learn the language and be explicit about root pitch declaration. `E4^5 = mtof(_)` can be a valid choice in some situations and there's no way to fit all imaginable extended FJS alternatives into a dropdown menu.

000masa000:
True. The "sharp" approach is "simpler", in a way. The eight original notes, from Bb to B, were Pythagorean. My only counter-argument here might be that using "sharp" defaults (as the composer Xenakis did) reinforces the very strongly ingrained sense that sharps and flats are somehow enharmonic equivalents (rather than being 12 fifths apart, and spanning a comma, as we generally prefer it).  Would it be an acceptable alternative to have a choice of b / # enharmonics on the black notes only? That would be convenient and elegant.

000masa000:
OK, sounds good!

PR #641, opened 2024-04-08
Remove dropdown for Pythagorean enharmonic

frostburn:
Use the sharp enharmonic by default (MIDI standard). Motivate the user to declare an explicit root pitch when using a black base note.

ref #640


frostburn:
Does what was agreed to in the issue. Merging.

PR #642, opened 2024-04-11
Added Multichannel-to-equave feature to support Lumatone and other multichannel many-note controllers

000masa000:
Added options in the MIDI tab to select multichannel mode, center channel (1-16), number of equaves (1-16), and equaves down (0-15). Added multichannel index offset calculation using mmod and user-selected parameters. If the index goes outside the 0-127 range, frequency is calculated directly using scale.getFrequency().
TODO: automatically select all channels when toggling multichannel mode.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "scale-workshop",
-  "version": "3.0.0-beta.10",
+  "version": "3.0.0-beta.11",
   "scripts": {
     "dev": "vite",
     "build": "run-p type-check \"build-only {@}\" --",
@@ -21,7 +21,7 @@
     "moment-of-symmetry": "^0.4.2",
     "pinia": "^2.1.7",
     "qs": "^6.12.0",
-    "sonic-weave": "github:xenharmonic-devs/sonic-weave#v0.0.11",
+    "sonic-weave": "github:xenharmonic-devs/sonic-weave#v0.0.12",
     "sw-synth": "^0.1.0",
     "temperaments": "^0.5.3",
     "vue": "^3.3.4",
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "scale-workshop",
-  "version": "3.0.0-beta.10",
+  "version": "3.0.0-beta.11",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "scale-workshop",
-      "version": "3.0.0-beta.10",
+      "version": "3.0.0-beta.11",
       "dependencies": {
         "isomorphic-qwerty": "^0.0.2",
         "ji-lattice": "^0.0.3",
@@ -14,7 +14,7 @@
         "moment-of-symmetry": "^0.4.2",
         "pinia": "^2.1.7",
         "qs": "^6.12.0",
-        "sonic-weave": "github:xenharmonic-devs/sonic-weave#v0.0.11",
+        "sonic-weave": "github:xenharmonic-devs/sonic-weave#v0.0.12",
         "sw-synth": "^0.1.0",
         "temperaments": "^0.5.3",
         "vue": "^3.3.4",
@@ -5414,8 +5414,8 @@
       }
     },
     "node_modules/sonic-weave": {
-      "version": "0.0.11",
-      "resolved": "git+ssh://git@github.com/xenharmonic-devs/sonic-weave.git#6e2e13671b926bc73998fe800862b8ce73b77892",
+      "version": "0.0.12",
+      "resolved": "git+ssh://git@github.com/xenharmonic-devs/sonic-weave.git#771e2e17fd1336a867d90db08b9b2625c448b251",
       "license": "MIT",
       "dependencies": {
         "moment-of-symmetry": "^0.4.2",
@@ -5425,7 +5425,7 @@
         "sonic-weave": "bin/sonic-weave.js"
       },
       "engines": {
-        "node": ">=10.6.0"
+        "node": ">=12.0.0"
       },
       "funding": {
         "type": "github",
--- a/src/stores/scale.ts
+++ b/src/stores/scale.ts
@@ -25,6 +25,7 @@ import {
   repr
 } from 'sonic-weave'
 import {
+  APP_TITLE,
   DEFAULT_NUMBER_OF_COMPONENTS,
   INTERVALS_12TET,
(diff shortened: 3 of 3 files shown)


package.json
```diff
@@ -28,7 +28,7 @@
     "vue-router": "^4.3.0",
     "webmidi": "^3.1.8",
     "xen-dev-utils": "^0.2.9",
-    "xen-midi": "^0.1.3"
+    "xen-midi": "github:xenharmonic-devs/xen-midi"
```
frostburn, line 31: Is `xen-midi` good enough now? I should publish current `main` then.

src/App.vue
```diff
@@ -71,11 +71,26 @@ function sendNoteOn(frequency: number, rawAttack: number) {
...
+    let offset =  mmod(channel - midi.multichannelCenter + midi.multichannelEquavesDown, midi.multichannelNumEquaves) - midi.multichannelEquavesDown
+    offset = offset * scale.scale.size
+    index = index + offset
+  }
+  
   if (rawAttack === undefined) {
     rawAttack = 80
   }
+
   let frequency = scale.frequencies[index]
```
frostburn, line 87: Now that the `index` can get out of range `scale.getFrequency(index)` should be used unconditionally. (It uses pre-calculated freqs when available.)

src/assets/main.css
```diff
@@ -139,6 +139,7 @@ ul.btn-group {
 }
 .control.checkbox-container {
   flex-flow: unset;
+  gap: 0.15rem 1rem;
```
frostburn, line 142: This affects all checkboxes in the application. Did you make sure we want this accross the board?

src/assets/main.css
```diff
@@ -149,6 +150,10 @@ ul.btn-group {
 .control.radio-group span label {
   font-weight: unset;
 }
+.control.select-group {
```
frostburn, line 153: "Select" is a particular HTML element. I don't see this being used with that tag. Try to find a synonym for this group type if possible.

package.json
000masa000: It is fine for multichannel use, please publish! I was thinking of adding another export MidiOutMTS which would give an alternative to multichannel PB using Midi Tuning Standard and sysex (opt in popup for browser users): The logic of retuning in Midi 2.0 is almost identical so the code will be re-useable when sw adds Midi 2.0 compatibility (roadmap 4.0???) Concept is to use the note numbers as slots in which the "actual" neareast MIDInote and needed PB are stored. Then one can simply cycle through the MIDI note slots and recover info from them for per-note stuff (NoteOff Velocity, Poly AT etc.)

src/App.vue
000masa000: OK, will do.

src/assets/main.css
000masa000: I did this because otherwise they cram together without spacing when placed side-by-side. It seems to look OK, what do you think?

src/assets/main.css
000masa000: OK, will do.

package.json
frostburn: Published. Please update to `"xen-midi": "^0.2.0"`

Please raise a dedicated issue in `xen-midi` about this MIDI 2.0 stuff, so your ideas won't get lost once this PR is closed.

src/assets/main.css
frostburn: It's too wide for my taste to be applied across the app.
A more conservative gap would be a nice addition, though. Some of the labels elements have extra spaces in them which looks bad in template code.

Beyond the scope of this PR anyway. Please include the changes you need under `<style scoped>` in MidiView.vue.

src/assets/main.css
000masa000: for now I replaced "select-group" with "dropdown-group" and also added "checkbox-group" to match the already existing "radio-group". Not sure but it seemed to make sense that a row of each of these HTML elements might need different spacing. Will move this to the new issue https://github.com/xenharmonic-devs/scale-workshop/issues/644.

src/App.vue
```diff
@@ -71,11 +71,21 @@ function sendNoteOn(frequency: number, rawAttack: number) {
   return off
 }
 
-function midiNoteOn(index: number, rawAttack?: number) {
+function midiNoteOn(index: number, rawAttack?: number, channel?: number) {
+
+  // in multichannel-to-equave mode calculate an offset based on the incoming channel
+  if (midi.multichannelToEquave && channel !== undefined) {
+    let offset =  mmod(channel - midi.multichannelCenter + midi.multichannelEquavesDown, midi.multichannelNumEquaves) - midi.multichannelEquavesDown
```
frostburn, line 78: `const offset; index += offset * scale.scale.size;` Might be nicer (prefer `const` over `let` when possible).

src/App.vue
```diff
@@ -71,11 +71,21 @@ function sendNoteOn(frequency: number, rawAttack: number) {
...
+    offset = offset * scale.scale.size
+    index = index + offset
+  }
+  
   if (rawAttack === undefined) {
     rawAttack = 80
   }
-  let frequency = scale.frequencies[index]
+
+  let frequency = scale.getFrequency[index]
```
frostburn, line 87: `(index)` instead of `[index]`.

src/assets/main.css
```diff
@@ -143,12 +143,19 @@ ul.btn-group {
 .control.checkbox-container label {
   font-weight: normal;
 }
+.control.checkbox-group {
+  flex-flow: unset;
+  gap: 0.15rem 1rem;
```
frostburn, line 148: We already discussed this. Don't modify global styles in this PR.

src/views/MidiView.vue
```diff
@@ -178,9 +178,24 @@ onUnmounted(() => {
...
+              <input type="checkbox" id="midi-velocity" v-model="midi.velocityOn" />
+              <label for="midi-velocity"> Use velocity </label>
+            </span>
+            <span>
+              <input type="checkbox" id="multichannel" v-model="midi.multichannelToEquave" />
+              <label for="multichannel"> Multichannel-to-equave </label>
+            </span>
+          </div>
+          <label>Settings for multichannel-to-equave mode</label>
+          <div class="control dropdown-group">
```
frostburn, line 192: There's nothing *dropdown* about this group of elements. See `btn-dropdown-group` on what that means.

frostburn:
The merge conflict needs to be resolved. Rebase on current `main`.

PR #654, opened 2024-04-16
fixed radio button spacing with css

000masa000:


src/assets/main.css
```diff
@@ -138,13 +152,13 @@ ul.btn-group {
   flex-grow: 0.25 !important;
 }
 .control.checkbox-container {
-  flex-flow: unset;
+  flex-flow: unset;;
 }
 .control.checkbox-container label {
```
frostburn, line 157: Why isn't this rule specific enough? Why the extra class?

frostburn:
Do all the labels. There's some you missed in e.g. ExpandScale.vue.

src/assets/main.css
000masa000: I will try your suggestion. I found I needed separate styling for the radio-group labels and the checkbox labels, so that visually they look similar.

000masa000: OK, it worked, thanks!

000masa000: wondering about the checkbox-group on MIDIView, would you prefer it on 2 lines (like elsewhere in the app)? I had to make a scoped style for it....

frostburn: I think MIDI view deserves dedicated styles. That many checkboxes in a row is once-in-an-app exception.

frostburn:
Code-wise looks good. Too many commits, but I can squash when merging on GitHub. Will review the UI after breakfast.

frostburn:
Good job! Squashing and merging.

PR #657, opened 2024-04-17
made multichannel-to-equave settings and color mapping dependent on mode

000masa000:
made options visible depending on multichannel-to-equave checkbox

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -72,8 +72,10 @@ function sendNoteOn(frequency: number, rawAttack: number) {
 }
 
 function midiNoteOn(index: number, rawAttack?: number, channel?: number) {
+  const multichannel = midi.multichannelToEquave
+
   // in multichannel-to-equave mode calculate an offset based on the incoming channel
-  if (midi.multichannelToEquave && channel !== undefined) {
+  if (multichannel && channel !== undefined) {
     let offset =
       mmod(
         channel - midi.multichannelCenter + midi.multichannelEquavesDown,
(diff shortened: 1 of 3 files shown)


frostburn:
If you're hiding the color mapping stuff there should be a corresponding bypass in App.vue to make sure multichannel mode doesn't behave strangely.

src/views/MidiView.vue
```diff
@@ -186,61 +186,65 @@ onUnmounted(() => {
...
-              <input
-                id="multichannel-equaves-down"
-                class="control"
-                type="number"
-                min="0"
-                max="15"
-                v-model="midi.multichannelEquavesDown"
-              />
+          </template>
+          <template v-if="!midi.multichannelToEquave">
```
frostburn, line 227: This could be a `v-else` but then the template becomes more rigid. Maybe two separate `v-if`s is better. Just wanted to point out the syntax.

000masa000: thanks, I noticed that possibility this morning while looking around other places where v-if pops up

000masa000:
2nd commit adds cahnges to App.vue and Midi Piano so chromatic mode behaviour reverts when in multichannel-to-equave mode

frostburn:
Missing bypass near `return (rawRelease?: number) => {` in App.vue.

src/App.vue
```diff
@@ -156,7 +158,7 @@ function midiNoteOn(index: number, rawAttack?: number, channel?: number) {
     if (!midi.velocityOn) {
       rawRelease = 80
     }
-    if (whiteMode === 'simple') {
+    if (whiteMode === 'simple' || multichannel) {
```
frostburn, line 161: I guess this is on me for writing bad code in the first place. Enough churn. I'll fix it on `main`.

PR #685, opened 2024-05-07
Update sonic-weave dependency

frostburn:
Remove obsolete items from the character palette.

SonicWeave changelog:
- Add support for unary √
- Remove superfluous accidentals
- Clarify relative interval qualities by making √2 a perfect interval
- Switch Latin -> Greek correspondence to be semioctave-alphabetical
- Drop absolute semiquartals from the grammar
- Implement modules
- Implement range/slice syntax for excluding the end point `[0 .. < 5]`
- Add an emoticon `[8>+%¶` for 256 Hz
- Fix formatting of non-standard vals
- Add amity inflection to Lumi's commas
- Implement range() like Python 2
- Make real values usable as counters


arseniiv:
> Add an emoticon [8>+%¶ for 256 Hz

wha 🤣

Issue #721, opened 2024-06-04
Await results of router.push


frostburn:
I'm unable to reproduce these issues myself, but obXen is reporting navigation bugs related to extraneous network activity when Programmatic Navigation is used.

frostburn:
I'm able to produce similar issues if assuptions related `to === from` are violated in `router.afterEach`.

PR #724, opened 2024-06-04
Decode URL only once

frostburn:
Rely on Pinia to maintain application state.
Make sure to await programmatic navigation.

ref #721


frostburn:
Somehow this breaks undo.

Issue #680, opened 2024-05-01
Smart lattice tab


frostburn:
Automatically infer the best lattice to show initially based on generation method and scale content. Do nothing if the user has touched lattice options during the session.

PR #727, opened 2024-06-05
Automatically infer good lattice configuration

frostburn:
Inspect scale contents to figure out good initial lattice configuration.

ref #680


frostburn:
Inference should be available at the push of a button later in the session too.

Inference should use the designated lattice intervals.

Issue #796, opened 2024-07-28
Interval matrix wolfy spelling issues


frostburn:
**0.5° 🍵** — _Today at 12:58 AM_

> any ideas about showing giant ratios in the interval matrix as tamer ratios up to comma(a) the user specifies? (searching just a bit around those commas, like adding one or two at best)
> 
> this can do something useful to Fokker blocks (insofar they get encoded in 1D beneficially) and my period-up-to-a-comma scales
> for example: https://scaleworkshop.plainsound.org/scale/tAjcMXJv6
> horrible ratios in places but the comma 42728167936000/42728053589487 can make them look nicer
> up to this tiny comma, the scale is MV4
> which also doesn’t get to be known if we don’t up the tolerance

PR #797, opened 2024-07-29
Apply CS margin to variety signature on Analysis tab.

frostburn:
Approximate Tenney-tall fractions on Analysis tab.

ref #796

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Change log
 
 ## 3.0.1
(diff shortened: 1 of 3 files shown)


arseniiv:
If it has a checkbox to toggle the old behavior, I’m in favor!

frostburn:
> If it has a checkbox to toggle the old behavior, I’m in favor!

It doesn't have a checkbox, but you can recreate the old behavior by setting the margin and max height to zero (which is interpreted as infinity for technical reasons).

arseniiv:
Fine with me, even more if there’s a small reminder in a tooltip or somewhere else.

Issue #803, opened 2024-08-11
Equal temperaments with cents equaves generate syntax errors


frostburn:
`1\5<1200.>` is illegal but `1\5 ed 1200.` is fine. Angle brackets vs. comparisons *sigh* :disappointed: ...

PR #808, opened 2024-08-14
Admit cents as the equave when generating equal temperaments

frostburn:
ref #803


frostburn:
Estimated time to review: 10 minutes

PR #822, opened 2024-11-08
Steps can be increamented with - Semitone, Tone, Octave. According to choice of user

sohamm20:
Screenshot 2024-11-08 at 9 35 24 PM

### DIFF
--- a/src/assets/main.css
+++ b/src/assets/main.css
@@ -126,7 +126,7 @@ ul.btn-group {
   display: flex;
   flex-flow: row wrap;
   align-items: stretch;
-  gap: 0.15rem;
+  gap: 0.15rem 0.15rem;
 }
 .control-group h2 {
   margin-bottom: unset;
--- a/src/assets/main.css
+++ b/src/assets/main.css
@@ -126,7 +126,7 @@ ul.btn-group {
   display: flex;
   flex-flow: row wrap;
   align-items: stretch;
-  gap: 0.15rem 0.15rem;
+  gap: 0.15rem;
 }
 .control-group h2 {
   margin-bottom: unset;
--- a/src/components/ScaleControls.vue
+++ b/src/components/ScaleControls.vue
@@ -5,6 +5,8 @@ import { ref } from 'vue'
 import ScaleRule from './ScaleRule.vue'
 import palette from '@/character-palette.json'
 
+const stepx = ref("1")
+
 const scale = useScaleStore()
 
 const updateScale = debounce(scale.computeScale)
@@ -53,13 +55,24 @@ defineExpose({ focus, clearPaletteInfo })
       <input
         id="base-midi-note"
         type="number"
-        step="1"
+        :step="stepx"
         v-model="scale.baseMidiNote"
         @input="updateScale()"
       />
       <span class="midi-name">{{
-        midiNoteNumberToName(scale.baseMidiNote, -1, scale.accidentalPreference)
-      }}</span>
+          midiNoteNumberToName(scale.baseMidiNote, -1, scale.accidentalPreference)
+        }}</span>
+
+      <span>
+        <label for="select-step">Step : </label>
+        <select id = "select-step" v-model="stepx">
+          <option value = "1">Semitone</option>
+          <option value = "2">Tone </option>
+          <option value = "12">Octave</option>
+        </select>
+
+      </span>
+
     </div>
 
     <div class="control">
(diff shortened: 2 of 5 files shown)


frostburn:
Was there community demand for this?
I personally think this just adds clutter.
We can have it, but please put it behind a preference that's off by default.

sohamm20:
Can you give suggestions on how to add it as a preference?

src/components/ScaleControls.vue
```diff
@@ -53,13 +55,24 @@ defineExpose({ focus, clearPaletteInfo })
...
         v-model="scale.baseMidiNote"
         @input="updateScale()"
       />
       <span class="midi-name">{{
-        midiNoteNumberToName(scale.baseMidiNote, -1, scale.accidentalPreference)
-      }}</span>
+          midiNoteNumberToName(scale.baseMidiNote, -1, scale.accidentalPreference)
+        }}</span>
+
+      <span>
```
frostburn, line 66: It's been a while since I've programmed Vue, but it's something along the lines of adding the new preference, `const state = useStateStore()` and replacing this line with:  `<span v-if="state.showMidiStepVariants">`.

sohamm20:
@frostburn Added it as a preferance. Can you take a look?

src/assets/main.css
```diff
@@ -126,7 +126,7 @@ ul.btn-group {
   display: flex;
   flex-flow: row wrap;
   align-items: stretch;
-  gap: 0.15rem;
+  gap: 0.15rem 0.15rem;
```
frostburn, line 129: Unnecessary change. If only one value is included, that value is used for both dimensions.

src/components/ScaleControls.vue
```diff
@@ -4,6 +4,10 @@ import { debounce, midiNoteNumberToName } from '@/utils'
 import { ref } from 'vue'
 import ScaleRule from './ScaleRule.vue'
 import palette from '@/character-palette.json'
+import { useStateStore } from '@/stores/state'
+const state = useStateStore()
+
+const stepx = ref("1")
```
frostburn, line 10: Uninformative variable name. What is it associated with?

src/stores/state.ts
```diff
@@ -54,6 +54,8 @@ export const useStateStore = defineStore('state', () => {
   // Debugging features.
   const debug = ref(storage.getItem('debug') === 'true')
 
+  const step = ref(storage.getItem('step') === 'true')
```
frostburn, line 57: Uninformative storage key. Step of what?

src/views/PreferencesView.vue
```diff
@@ -54,6 +54,10 @@ const scale = useScaleStore()
             <input id="debug" type="checkbox" v-model="state.debug" />
             <label for="debug" class="right-of-checkbox">Enable debugging features</label>
           </div>
+          <div class="control checkbox-container">
+            <input id="custom-step" type="checkbox" v-model="state.step" />
+            <label for="custom-step" class="right-of-checkbox">Enable Custom Step Increment</label>
```
frostburn, line 59: Not enough information for the user. Which step increment is this?

frostburn:
In addition to the step increment. Setting the property to anything other than 1 also fixes the "root" value. Is it intentional behavior that e.g. 61 is an invalid value when the *Tone* size is selected? Same goes for the *Octave* option: Are 0, 12, 24, etc. exactly the values you want?

frostburn:
The motivation of this PR still escapes me. Can you elaborate on why you'd want this?

frostburn:
If you wish to, remember to add your name to the *Contributors* list on the about page.

sohamm20:
I often need to switch between lower and higher octaves while staying on the same note. Normally, this would require pressing up or down 12 times, but this option makes the process much more convenient.

sohamm20:
> In addition to the step increment. Setting the property to anything other than 1 also fixes the "root" value. Is it intentional behavior that e.g. 61 is an invalid value when the _Tone_ size is selected? Same goes for the _Octave_ option: Are 0, 12, 24, etc. exactly the values you want?

Could we grey out the button if selecting it would result in an invalid root value?

frostburn:
> I often need to switch between lower and higher octaves while staying on the same note. Normally, this would require pressing up or down 12 times, but this option makes the process much more convenient.

That's not what MIDI base note is for. It's more relevant for exporters.

See keyboard shortcuts *equave shift up/down* for what you're doing.
image

frostburn:
Can you try the equave shift keys to see if they already do what this PR is trying to achieve?

If you're on mobile or otherwise do not have access to a numpad section on your keyboard I think you should replace this PR with a one that makes *equave shift* into an on-screen button.

sohamm20:
Sure, I will try

Issue #379, opened 2022-12-10
Show comma relative to period when generating Rank-2


frostburn:
[Weltfried](https://www.reddit.com/user/Weltfried/) [on reddit](https://www.reddit.com/r/musictheory/comments/zexkrm/comment/izixtwl/?utm_source=reddit&utm_medium=web2x&context=3)

> It would be nice if the Rank-2 temperament Generator dialogue told you the comma relative to the period when picking a MOS. That's useful information for tempering and would save the trouble of calculating it by hand. (Append generator one more time after all have been appended in the same direction and calculate interval relative to period). Maybe that feature is already present, but I couldn't find it.

PR #836, opened 2025-01-05
Display hardness and chroma alongside the MOS pattern in Rank 2

frostburn:
Update info section of the Rank 2 generation modal.

ref #379


frostburn:
image
The new stuff is L/s and L-s at the bottom.

PR #850, opened 2025-01-25
Move non-algorithmic aperiodic timbres to the JSON file

frostburn:


frostburn:
Self-reviewing: Looks good to me. Will merge later today if nothing comes up.

arseniiv (approved):
Looks reasonable, I looked at the branch yesterday and found nothing wrong. Data structure naming is the best I could think of.

PR #853, opened 2025-02-06
Add musical notation

kfxs:
This PR, adds a staff notation widget on every tab.....


src/stores/state.ts
```diff
@@ -3,7 +3,12 @@ import { defineStore } from 'pinia'
 import { UNIX_NEWLINE } from '@/constants'
 import { syncValues } from '@/utils'
 
+
+
 export const useStateStore = defineStore('state', () => {
+  // Nonpersistent state of the application
+  const scoreChord = ref<string[]>([])
```
frostburn, line 10: This should be a `reactive`. You're pushing to it elsewhere which breaks Vue.

src/App.vue
```diff
@@ -228,12 +230,31 @@ watch(
   }
 )
 
+
+// === Score Chords functions ===
```
frostburn, line 234: These are strongly associated with `state`. Makes more sense to move them to that file.

src/App.vue
```diff
@@ -228,12 +230,31 @@ watch(
...
 
+
+// === Score Chords functions === 
+function pushSymbolToChord(index: number){
+  state.scoreChord.push(scale.symbols[index])
+}
+
+function pullSymbolfromChord(index: number){
+  const item = state.scoreChord.indexOf(scale.symbols[index])
+  state.scoreChord.splice(item, 1)
```
frostburn, line 241: Does `scoreChord` really need to be an `Array`? I don't think the order matters in a vertically stacked chord so a `Set` makes more sense semantically.

src/App.vue
frostburn: Scratch that. You're mixing `scale` and `state`. This is fine here.

src/components/ScaleControls.vue
```diff
@@ -5,14 +5,33 @@ import { ref } from 'vue'
...
 
+
+function clearUserNotation(){
+  scale.userNotation = ''
+  updateScale()
+}
+
+function increaseOttava() {
+  if(scale.ottava < 3){
+    scale.ottava ++
```
frostburn, line 23: Please do `npm run format`.

src/components/ScoreView.vue
```diff
@@ -0,0 +1,124 @@
...
+import Vex from 'vexflow';
+
+
+const state = useStateStore()
+const { Factory, BarlineType } = Vex.Flow;
+
+function isTrebleClef(chord: string | undefined){
+    if(!chord) return false
+
+    const number = Number(chord.charAt(chord.length-1))
```
frostburn, line 13: This fails for octaves larger than 9. Also `parseInt(str, 10)` is slightly more stable than `Number`.

src/components/ScoreView.vue
```diff
@@ -0,0 +1,124 @@
...
+    if(number >= 4) return true
```
frostburn, line 14: I personally prefer a simple `return (number >= 4)`.

src/components/ScoreView.vue
```diff
@@ -0,0 +1,124 @@
...
+    const vf = new Factory({
+        renderer: { elementId: 'chordView', width: 160, height: 160 },
+    })
+    vf.context.scale(0.75, 0.75)
+
+    let chordArrayTreble = []
+    let chordArrayBass = []
+    for(const item of chordArray){
+        if(isTrebleClef(item)) chordArrayTreble.push(item)
+        if(isBassClef(item)) chordArrayBass.push(item)
```
frostburn, line 42: Aren't the cleffs mutually exclusive? Just `else` here.

kfxs: No, the are not. We have to systems. One for the treble and another for the bass. So we have two clefs.

src/components/TuningTable.vue
```diff
@@ -45,6 +47,7 @@ const rows = computed(() => {
         <th>Cents</th>
         <th>Ratio</th>
         <th>Label</th>
+        <th>Symbol</th>
```
frostburn, line 50: This column should be conditioned on `state.showMusicalScore`.

frostburn:
Feel free to add your details on the Hall of Fame on the about page.

A CHANGELOG entry would be nice too, but this feature likely requires some tweaking and additional support on `sw-server` side, so I can add that later. (Eventually on `sonic-weave` side too, but I think we can release a beta preview using the textarea.)

src/components/ScoreView.vue
frostburn: Two systems yes, but an item is never pushed to two places at the same time, right?

kfxs: Ah, now I get it. Ok!

kfxs:
Hi @frostburn; can you please give a look to my changes?

src/views/AboutView.vue
```diff
@@ -64,7 +64,8 @@ const tagline = computed(() => TAGLINES[Math.floor(Math.random() * TAGLINES.leng
         Richie Greene - <i>quality assurance</i> <br />
         Akselai - <i>quality assurance</i> <br />
         <span class="spoob">Spoob<img src="@/assets/img/spoob.png" /></span> -
```
frostburn, line 66: Spoob isn't a real person and should probably stay at the bottom. Gotta admit: It's funnier this way around.

src/components/ScaleControls.vue
```diff
@@ -123,6 +140,28 @@ defineExpose({ focus, clearPaletteInfo })
     </div>
     <p class="info" v-html="paletteInfo"></p>
   </div>
+
+  <div class="control-group">
```
frostburn, line 144: This section should be conditional on `state.showMusicalScore`.

src/views/PreferencesView.vue
```diff
@@ -46,6 +46,10 @@ const scale = useScaleStore()
             <input id="virtual-qwerty" type="checkbox" v-model="state.showVirtualQwerty" />
             <label for="virtual-qwerty" class="right-of-checkbox">Virtual QWERTY in top menu</label>
           </div>
+          <div class="control checkbox-container">
+            <input id="show-score" type="checkbox" v-model="state.showMusicalScore" />
+            <label for="show-Score">Show Musical Score</label>
```
frostburn, line 51: How about "Show staff notation widget" instead? I'm not sure what's the best label to use for the average users.

frostburn:
Clusters of chromatic notes render on top of each other.
image

frostburn:
Clicking on the "bassa" or "alta" buttons while a chord is playing leaves permanent notes on the widget.

frostburn:
Remember to `npm run test:unit` and to fix that failing unit test.

src/views/AboutView.vue
kfxs: Let's leve it like this then.

kfxs:
RE: Clusters of chromatic notes render on top of each other.

This is because of the way vexflow handles the chords and accidental. The issue is directly on vexflow.
 At the moment, I don't think there is a quick solution to that.

src/views/PreferencesView.vue
kfxs: Probably, is better. I'm not a native english speaker. I will cange it.

kfxs:
RE: Clusters of chromatic notes render on top of each other.

in a way, the bug seems to be related somehow with the register:
imagen

imagen

frostburn:
RE: Chromatic  clusters

Have you checked if Vexflow expects chords to be in ascending order to render properly? Maybe it broke when we converted stuff to use `Set`.

kfxs:
I'm just checking my old version and the issue is there:
imagen
so, it also happens with the Array.

kfxs:
all changes were pushed, except for the cluster issue.....

src/components/ScaleControls.vue
```diff
@@ -1,18 +1,65 @@
...
-
 const paletteInfo = ref('')
 
+function clearUserNotation() {
+  scale.userNotation = ''
+  updateScale()
+}
+
+function updateScoreChordOctave(step: number) {
+  console.log('--------------------')
```
frostburn, line 22: Is this PR a draft or intended for production? Please no spam to users' consoles in production or even staging.

frostburn:
Looking good in general. :+1:

I continue to be busy with real life stuff in the coming days and this feature will have to wait in limbo until I can decide how to handle this on `sw-server` side and if it's worth it to release a production build with a separate text area for chord symbols. I personally like inline syntax so there would need to be a data migration once that text area is removed.

src/stores/scale.ts
```diff
@@ -561,6 +624,10 @@ export const useScaleStore = defineStore('scale', () => {
     latticeEquave,
     colors,
     labels,
+    userNotation, //----(proof of concept)
```
frostburn, line 627: You can't just add new stuff to `LIVE_STATE` without taking versions into account.
After this update old scales stop working. To reproduce save a scale on `main` and try to load it on `add-musical-notation`.
Please add versioning so that old scales still load.
Please implement unit tests to confirm that old and new scales load correctly.

src/App.vue
```diff
@@ -228,12 +230,28 @@ watch(
   }
 )
 
+// === Score Chords functions ===
+function pushSymbolToChord(index: number) {
```
frostburn, line 234: Pushing and pulling symbols from `state.scoreChord` is not robust.
To see this press shift and a key on the QWERTY keyboard. This triggers sustain. Edit the entry corresponding to the sustained note. Repress the key to release sustain. `state.scoreChord.delete(item)` now does nothing and the previous symbol is permanently displayed in the widget.

Ideally the chord shown in the widget would be a `computed` of the notes currently being played.

frostburn:
You should probably show a warning when the number of symbols differs from scale size.

src/components/ScaleControls.vue
```diff
@@ -1,18 +1,61 @@
...
+const state = useStateStore()
 const scale = useScaleStore()
-
 const updateScale = debounce(scale.computeScale)
 
 const sourceEditor = ref<HTMLTextAreaElement | null>(null)
-
 const paletteInfo = ref('')
 
+function clearUserNotation() {
```
frostburn, line 16: Should there be a method to restore the default symbols too?

src/components/ScaleControls.vue
```diff
@@ -197,4 +262,36 @@ p.warning {
...
   overflow-y: auto;
 }
+
+.ottava-label {
+  color: var(--color-accent-mute);
+}
+.ottava-value {
+  padding-top: 2px;
+  padding-bottom: 2px;
+  padding-left: 8px;
```
frostburn, line 272: These should probably be in `rem` in light of #859. I've used the wrong units on `main`, sorry about that.

src/components/ScaleControls.vue
```diff
@@ -123,6 +166,28 @@ defineExpose({ focus, clearPaletteInfo })
     </div>
     <p class="info" v-html="paletteInfo"></p>
   </div>
+
+  <div :class="['control-group', state.showMusicalScore ? 'visible' : 'hidden']">
```
frostburn, line 170: `v-if` or `v-show` instead of magic classes, please.

src/components/TuningTable.vue
```diff
@@ -27,6 +31,7 @@ const rows = computed(() => {
       cents,
       ratio: ratio,
       label: props.labels[mmod(index - 1, props.labels.length)],
+      symbol: props.symbols[i] || '',
```
frostburn, line 34: Nullish coalescing `??` is better for default values.

src/components/TuningTable.vue
```diff
@@ -45,6 +50,7 @@ const rows = computed(() => {
         <th>Cents</th>
         <th>Ratio</th>
         <th>Label</th>
+        <th :class="{ hidden: !state.showMusicalScore }">Symbol</th>
```
frostburn, line 53: `v-if` or `v-show` instead.

src/components/VirtualKeyboard.vue
```diff
@@ -66,7 +66,6 @@ const virtualKeys = computed(() => {
 
 const isMousePressed = ref(false)
 </script>
-
```
frostburn, line 69: Irrelevant change.

src/stores/scale.ts
```diff
@@ -65,6 +65,59 @@ function defaultLabels(base: number, accidentalStyle: AccidentalStyle) {
...
+  if (noteNames.length == 0) return symbolTable
+
+  let octave
+  if (noteNames.length == 1) {
+    octave = 4 + ottava
+  } else {
+    octave = 4 - Math.round(baseMidiNote / noteNames.length) + ottava
+  }
+
+  for (let i = 0; i <= midiRange; i++) {
```
frostburn, line 98: `i < NUMBER_OF_NOTES` instead.

kfxs: This was replaced. I didn't realized before there was a NUMBER_OF_NOTES constant (very handy).
However, from a user experience perspective I still find a little bit annoying not to show notes beyond the midi range. Moreover when the app produces sounds for those notes.

src/components/VirtualKeyboard.vue
kfxs: Do you mean the empty line? If that's the case I totally agree.

kfxs:
> You should probably show a warning when the number of symbols differs from scale size.

I added this warning:

imagen

Nevertheless, as a composer I would like to view also notes > NUMBER_OF_NOTES.
It would be nice to think on a solution for cases like this.

src/stores/scale.ts
kfxs: Well, this is definitely not an easy task for me. So I will write down what my intuitions suggest me:

For some reason, I have the feeling that the GET call to BE is happening here: https://github.com/kfxs/scale-workshop/blob/add-musical-notation/src/views/LoadScaleView.vue#L45

It seems to me that this does a fetch on  `/tmp/sw/scale/scale_id.json` on the server side. However, after giving a look to that  file I can not find any information regarding the version of SW3 (maybe I'm wrong and the file is just too long for my understanding).

However, there is indeed valuable information in `/tmp/sw/envelope/scale_id.envelope.json`.  These are a couple of examples:
imagen

imagen


It seems that `clientVersion` stores the version of SW3 in which the scale was created....
The problem now is that I can't find in the code any call to  `/tmp/sw/envelope/scale_id.envelope.json`.

So, what comes to my mind now is doing a GET call to that URL (envelope) just before https://github.com/kfxs/scale-workshop/blob/add-musical-notation/src/views/LoadScaleView.vue#L45 and then use the information of `clientVersion` to make a flag that handles the inclusion of `userNotation` , `noteNames` ,  `symbols` and  `ottava`  into `LIVE_STATE`.


Does it make any sense to you @frostburn ?

frostburn: I think we can duplicate `clientVersion` in the scale JSON or add a new `scaleVersion` field or something. The envelopes are just for tracking in case we ever need to ban some IPs or something. Making them downloadable would be a huge security issue.

kfxs: Good idea; and since  old scales will not have `clientVersion` it will be easy to know that were created in previous versions.

kfxs: well, cause I'm not very sure how to handle this in a proper way I have created a separated branch: https://github.com/kfxs/scale-workshop/tree/add-musical-notation-control-versions

Since I have no experience with Github; I'm not really sure how to handle this. I mean, I guess I can open a new PR (draft) to my current branch, but to be honest I have no idea....

@frostburn can you please suggest me how to proceed?

frostburn: I don't think we want this stuff in `main` as is. Try something like `3.1.0-notation-preview`. We're going to let people play around with the widget and give feedback. Later we will repurpose the work done here for the inline symbols and merge that to `main` when it's ready.

I'll add your branch in this repository when everything is ready for deploying to staging.

As for GitHub stuff, here are the alternatives:
1. Merge `add-musical-notation-control-versions` to `add-musical-notation` on your side. Things will change here.
2. Change where this PR is pointing to.
3. Make a new PR for the new branch and close this one.

kfxs: Where does this PR should be pointing to?

frostburn: I can't seem to find the option to do that in my pull request so point 2 is invalid, sorry.

kfxs: Well, I think I fix (sort of) the issue with older version scales. The approach was to simple add to the JSON the missing fields that were not in the older version scales. I made a test local, creating a scale on `main` and then used that URL to load the scale on `add-musical-notation`. It seems to me that is working.

kfxs: Regarding the unit test. To be honest, I have no idea of how to do this. Any advice?

src/stores/__tests__/scale.spec.ts
```diff
@@ -4,7 +4,7 @@ import { useScaleStore } from '../scale'
...
...ency":261.63,"autoFrequency":true,"autoColors":"silver","sourceText":"tet(5)","userNotation":"C\nC#\nD\nD#\nE\nF\nF#\nG\nG#\nA\nA#\nB","noteNames":["C","C#","D","D#","E","F","F#","G","G#","A","A#","B"],"symbols":["C-1","C#-1","D-1","D#-1","E-1","F-1","F#-1","G-1","G#-1","A-1","A#-1","B-1","C0","C#0","D0","D#0","E0","F0","F#0","G0","G#0","A0","A#0","B0","C1","C#1","D1","D#1","E1","F1","F#1","G1","G#1","A1","A#1","B1","C2","C#2","D2","D#2","E2","F2","F#2","G2","G#2","A2","A#2","B2","C3","C#3","D3","D#3","E3","F3","F#3","G3","G#3","A3","A#3","B3","C4","C#4","D4","D#4","E4","F4","F#4","G4","G#4","A4","A#4","B4","C5","C#5","D5","D#5","E5","F5","F#5","G5","G#5","A5","A#5","B5","C6","C#6","D6","D#6","E6","F6","F#6","G6","G#6","A6","A#6","B6","C7","C#7","D7","D#7","E7","F7","F#7","G7","G#7","A7","A#7","B7","C8","C#8","D8","D#8","E8","F8","F#8","G8","G#8","A8","A#8","B8","C9","C#9","D9","D#9","E9","F9","F#9","G9"],"ottava":0,"error":"","warning":"","isomorphicVertical":5,"isomorphicHorizontal":1,"keyboardMode":"isomorphic","equaveShift":0,"degreeShift":0,"pianoMode":"Asdf","accidentalColor":"black","lowAccidentalColor":"maroon","middleAccidentalColor":"navy","highAccidentalColor":"indigo"}`
```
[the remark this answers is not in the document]
kfxs: Hi @frostburn; I tried to create a new test for scale.legacy.specs.ts  based on scale.spec.ts; the test runs but since I have no experience at all on test units, I'm not really sure if what I'm doing is correct or not

src/stores/__tests__/scale.legacy.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+
+describe('Scale store', () => {
+    beforeEach(() => {
+      // creates a fresh pinia and makes it active
+      // so it's automatically picked up by any useStore() call
+      // without having to pass it to it: `useStore(pinia)`
+      setActivePinia(createPinia())
+    })
+  
+    it('serializes 5edo', () => {
```
frostburn, line 29: You made a new version. It cannot serialize to the legacy version. Remove this test.

src/stores/__tests__/scale.legacy.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+      const scale = useScaleStore()
+      scale.sourceText = 'tet(5)'
+      scale.computeScale()
+      const legacyData = JSON.stringify(scale)
+      expect(legacyData).toBe(LEGACY_SCALE)
+    })
+  
+    it('deserializes 5edo', () => {
+      const data = JSON.parse(LEGACY_SCALE, (key: string, value: any) =>
```
frostburn, line 38: You're reviving monkey-patched data. The server will not serve that. It will serve old data as is. (I really don't want to write a data migration for this.) Your system needs to work directly with `LEGACY_DATA`.

src/stores/__tests__/scale.legacy.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+      expect(centss[3]).toBeCloseTo(720)
+      expect(centss[4]).toBeCloseTo(960)
+      expect(centss[5]).toBeCloseTo(1200)
+      expect(scale.latticeIntervals).toHaveLength(5)
+    })
+  
+    it('has a cap on the number of shareable intervals', () => {
+      const scale = useScaleStore()
+      scale.sourceText = '500::1000'
+      scale.computeScale()
```
frostburn, line 59: `computeScale` uses new code so it's irrelevant inside a legacy compatibility test.

src/stores/__tests__/scale.legacy.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+      expect(centss[0]).toBeCloseTo(0)
+      expect(centss[1]).toBeCloseTo(240)
+      expect(centss[2]).toBeCloseTo(480)
+      expect(centss[3]).toBeCloseTo(720)
+      expect(centss[4]).toBeCloseTo(960)
+      expect(centss[5]).toBeCloseTo(1200)
+      expect(scale.latticeIntervals).toHaveLength(5)
+    })
+  
+    it('has a cap on the number of shareable intervals', () => {
```
frostburn, line 56: Just remove this test case.

vite.config.ts
```diff
@@ -12,5 +14,8 @@ export default defineConfig({
     alias: {
       '@': fileURLToPath(new URL('./src', import.meta.url))
     }
+  },
+  define: {
+    'import.meta.env.PACKAGE_VERSION': JSON.stringify(version)
```
frostburn, line 19: What is this? `utils.makeEnvelope` already worked fine without meta env tricks as far as I know. Why is this necessary?

src/stores/scale.ts
```diff
@@ -124,6 +178,14 @@ export const useScaleStore = defineStore('scale', () => {
   const latticeEquave = ref<Interval | undefined>(undefined)
   const colors = ref(defaultColors(baseMidiNote.value))
   const labels = ref(defaultLabels(baseMidiNote.value, accidentalPreference.value))
+
+  // Notation widget
+  const swScaleVersion = ref(makeEnvelope().version)
```
frostburn, line 183: Just import `version` like `utils` does. No need for `ref` either. It won't change during a user session.
(I meant `makeEnvelope` as an example on how to do things, not as a source of truth.)

kfxs: Ok

kfxs: Now json file for new scales looks like this:
imagen

PR #916, opened 2026-04-10
Keep the Scale View alive permanently

### LINKED ISSUE #865: Regression: Ctrl + Z not working
**Finky** — _7:47 AM_

> problem I have with SW3 is that I can't go back in my history to previous versions of the scale
> so if I go to the lattice or something and then back to the scale builder, ctrl+z no longer does anything

frostburn:
Fixes a Ctrl+Z undo regression.

ref #865


frostburn:
Tests are broken but it's not the fault of these changes.

PR #919, opened 2026-04-12
Add delay presets to Scale Workshop

tristanbay:


src/views/SynthView.vue
```diff
@@ -79,6 +79,45 @@ function presetLong() {
...
+}
+
+function dlyPresetAmbient() {
+  audio.pingPongDelayTime = 0.019
+  audio.pingPongFeedback = 0.625
+  audio.pingPongSeparation = 0.8
+  audio.pingPongGain = 0.6
+}
+
+function dlyPresetOff() {
```
frostburn, line 117: Strictly not a preset, but good to have. Let's iterate in a separate issue.

frostburn (approved):

Issue #917, opened 2026-04-12
Type indexing error in SynthView.vue


tristanbay:
> I get an error on this line in VSCodium:
> 
> 
> 
> Element implicitly has an 'any' type because expression of type '"" | "deactivationCode" | "equaveUpCode" | "equaveDownCode" | "degreeUpCode" | "degreeDownCode"' can't be used to index type '_StoreWithState<"state", Pick<{ heldNotes: Reactive<Map<number, number>>; typingActive: Ref<boolean, boolean>; latticeType: Ref<"ji" | "et" | "cycles" | "3d" | "auto", "ji" | "et" | "cycles" | "3d" | "auto">; ... 23 more ...; fromJSON: (data: any) => void; }, "newline" | ... 23 more ... | "colorScheme">, Pick<...>, ...'.
> 
>   Property '' does not exist on type '_StoreWithState<"state", Pick<{ heldNotes: Reactive<Map<number, number>>; typingActive: Ref<boolean, boolean>; latticeType: Ref<"ji" | "et" | "cycles" | "3d" | "auto", "ji" | "et" | "cycles" | "3d" | "auto">; ... 23 more ...; fromJSON: (data: any) => void; }, "newline" | ... 23 more ... | "colorScheme">, Pick<...>, ...'. 

 _Originally posted by @tristanbay in [4ff2858](https://github.com/xenharmonic-devs/scale-workshop/commit/4ff2858509e90aaafce7333b2fb1d2b808740569#r182174782)_

PR #923, opened 2026-04-12
Fix remappable key indexing type narrowing

frostburn:


frostburn:
@tristanbay does this fix #917 ? (I refuse to run VSCode on the host machine until the current security storm settles.)

tristanbay:
Which security storm? I haven't heard of it. Does it affect VSCodium as well or is it a Microsoft-specific thing?

tristanbay:
And yes, it fixes #917. The error disappears and unit and e2e tests all pass.

frostburn:
> Which security storm? I haven't heard of it. Does it affect VSCodium as well or is it a Microsoft-specific thing?

On top of my head I can recall the SHA1-Hulud worm and the recent Axios hack that smuggled a `plain-crypto` package that run a malicious post-install hook spreading backdoors and other terrible stuff.

With AI agents getting smarter every month it's expected that the jailbroken versions will write an insane amount of software exploits soon. If Claude Mythos hacking open BDS is real and not just a PR stunt we're heading into strange times indeed.

I'm trying to run everything in containers and virtual machines these days... Those sandboxes can be broken out of, but at least it stops the simplest worms from spreading.

Issue #922, opened 2026-04-12
Visually Indicate Special Delay Preset "Off" behavior


frostburn:
The new "Off" preset doesn't set every value. Make it visually distinct from the other preset buttons or at least group it a bit further away.

frostburn:
@tristanbay What do you think? Does this sound reasonable?

tristanbay:
That sounds fine to me

PR #924, opened 2026-04-12
Separate and highlight off button for delay

tristanbay:
Potential solution for issue #922

I screwed up a little by syncing the main branch of my fork with the upstream after creating the new branch and then merging those changes into the new branch, which is why the same commits are also listed in the PR. It doesn't seem to affect the files any further though.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,7 @@ coverage
 *.njsproj
 *.sln
 *.sw?
+*.code-workspace
 
 # Production secrets
 .env.production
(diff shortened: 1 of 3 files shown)


tristanbay:
[This commit](https://github.com/xenharmonic-devs/scale-workshop/pull/924/commits/639f683fda552c6d52b5b3c712eccdc30c09375f) is the only relevant one. All of the others are redundant but don't change anything as far as I know.

PR #935, opened 2026-04-16
Fix center tuning table row scrolling into view

### LINKED ISSUE #934: RC Regression: Center row no longer focused on app load

frostburn:
Monitor ScaleView visibility and trigger scroll as necessary.

ref #934


frostburn:
Don't merge yet. We're deploying QA patches off-trunk.

PR #953, opened 2026-04-22
Make browser tabs resumable

### LINKED ISSUE #806: UX regression: Option to restore last session
**osmium** — _Today at 4:18 AM_

> im not sure if i should report this so apologies if not but it seems that the scale data isnt present on this version ? https://scaleworkshop.plainsound.org/?version=3.0.1 like i was experimentally adding notes to a scale then i accidentally closed the tab and went into my history but it's all blank

**frostburn** — _Today at 9:16 AM_

> That's just how it is in Scale Workshop 3. Stuff is stored on the server now. If you don't save an URL it's gone. We could try to store the last undo snapshot in local storage though...

frostburn:
Update window location hash with scale id when available. Store sessions for later.

ref #806


frostburn:
Still needs to upload scales when the browser tab becomes inactive.

frostburn:
Still needs to clear the id when it becomes stale in other stores too.

frostburn:
Deployed to QA for testing.

PR #986, opened 2026-07-18
Similar scales cleanup

### LINKED ISSUE #985: Similar scales cleanup
* Move *Similar scales* template code and styles to a dedicated Vue component.
* Remove `similar-scales.ts`, re-implement in the npm package and import from there.

narenratan:
- Move similar scales logic to `scale-library` npm package and import it from there
- Move similar scales template code and styles to their own component

Resolves #985.

I created a corresponding [scale-library PR](https://github.com/narenratan/scale-library/pull/2) in case it's of interest; it moves the similar scales code and sets up the packaging and testing config.


package.json
```diff
@@ -20,7 +20,7 @@
     "format": "prettier --write --experimental-cli src/"
   },
   "dependencies": {
-    "@narenratan/scale-library": "1.3.0",
+    "@narenratan/scale-library": "1.3.1",
```
frostburn, line 23: The PRs are looking good. My only nitpick is this patch version bump. Adding library functions is at least worth a minor bump. I do appreciate that you're only publishing a small part of a larger system so we can let it slide.

frostburn (approved):

package.json
narenratan: Thanks for having a look! Agreed it should really be a minor version bump; I thought I'd try and keep the main `scale-library` release, the npm package, and the Python package versions consistent (the main `scale-library` release is at `1.3.0` at the moment).

I've merged the `scale-library` PR and the new `scale-library` npm package version is published too.

frostburn: Thanks, I'll merge tomorrow morning.

PR #995, opened 2026-09-13
Replace the harmonium timbre with a measured spectrum

sohamm20:
The harmonium waveform was a hand-tuned approximation: a PWM square wave at duty cycle 0.43, shaped by odd-harmonic bias and brightness trim, then smeared across seven detuned copies at arbitrary offsets. It did not sound much like a harmonium.

This replaces it with a spectrum measured from a recording of a real instrument. Analysing one sustained C3 note showed it is not one rank but three sounding together:

  - an 8' bass reed                          -> ratio 1
  - a 4' octave reed 1202.48 cents above it  -> ratio 2.002862, a near-
    octave 2.48 cents wide, so the two ranks beat at 0.375 Hz; this is
    where the shimmer comes from, and why a pure 2:1 sounds lifeless
  - a third quiet rank 153.18 cents flat     -> ratio 0.915321, beating
    against the 8' at 11.1 Hz, heard as crispness rather than as pitch

Amplitudes came from short-window tracking; a long-window fit badly underestimates high partials because the instrument's own pitch wobble destroys their phase coherence. The 8' rank's odd partials run about 9 dB above its even ones, the signature of a free reed beating against its frame.

The result is 139 partials against roughly 480 before, so it is also lighter on the aperiodic voice budget.

Selecting the harmonium now also applies a matching envelope: ~120 ms attack (the measured reed speech time), no decay, full sustain, 100 ms release. A harmonium is blown rather than struck, and the app's default envelope decaying to 80% reads as a struck instrument and undoes the point of the measured spectrum.


sohamm20:
@forrcaho please review

src/stores/audio.ts
```diff
@@ -327,6 +328,16 @@ export const useAudioStore = defineStore<'audio', AudioStore>('audio', () => {
     if (APERIODIC_WAVEFORMS.includes(newValue)) {
       aperiodicVoiceParams.aperiodicWave = APERIODIC_WAVES[newValue].value
     }
+    if (newValue === 'harmonium') {
```
frostburn: Don't do this. The user can select an envelope preset. You can add a new one if "Organ" and "Pad" don't serve you.

frostburn:
Sounds good but it's too loud compared to a plain semisine.

sohamm20:
@frostburn fixed comments

frostburn (approved):
I'll deploy to QA for community feedback tomorrow.

frostburn:
QA: https://sw3.lumipakkanen.com/?version=3.5.4-rc1

